PR #5, opened 2023-03-31
feat(metrics): add Prometheus collector

SOF3:


gary-lgy (changes requested):
Let's fix the issues from the previous discussion before merging this:
1. Prometheus metric names cannot contain dots
2. Prometheus doesn't have a `Rate` metric

PR #7, opened 2023-03-31
refactor: clean up FrameworkHandle code, removed unused fields

SOF3:


pkg/controllers/scheduler/handle.go
```diff
@@ -0,0 +1,38 @@
...
+// The design and implementation of the scheduler is heavily inspired by kube-scheduler and karmada-scheduler. Kudos!
+
+package scheduler
+
+import (
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework"
+	"k8s.io/client-go/dynamic"
+)
+
+func (s *Scheduler) buildFrameworkHandle() framework.Handle {
```
gary-lgy, line 26: nit: why is this a method on `Scheduler`?

SOF3: @gary-lgy because it is an uti to read many misc fields from the `Scheduler`. The whole point of this method is to translate fields from `*Scheduler` to the methods in `framework.Handle`.

gary-lgy: Just curious, why not define the constructor as a standalone function? If we define it as a method on scheduler, we cannot instantiate a `Handle` outside the scheduler.

SOF3: Actually I am considering if I could just make the `*handle` contain a field that references `*Scheduler`, then just multiplex to the fields in the impl methods.

SOF3: I was concerned if there is a use case in the future (especially after @limhawjia implements dynamic scheduling profiles) where the handle fields would need to be dynamically updated and can only be fetched during actual use.

SOF3: @gary-lgy bump

gary-lgy (approved):

PR #14, opened 2023-04-04
feat: collect cluster resources without caching pods

limhawjia:
**Overview**

Collecting member cluster resources is required for resource-based scheduling plugins, such as ClusterResourcesFit, ClusterResourcesBalancedAllocation, etc.

Currently, we collect the available and allocatable member cluster resources by maintaining a Node and Pod informer for each member cluster. The number of schedulable nodes and the total allocatable resources is obtained from the Node informer. The total requested resources (which is subtracted from the total allocatable resources to get the total available resources) is obtained from the Pod informer.

This naturally does not scale well as informers maintain a copy of each object in memory.

This PR introduces a new mechanism for collecting cluster resources that **does not require the caching of pod objects**. With this new mechanism, we no longer require the use of a pod informer to collect the total requested resources in a cluster.

This new mechanism works based on the premise that:

1. Pod resource requests cannot be changed
2. List pagination is consistent and returns a single list resource version

**How the new mechanism works**

1. We **start with a paginated list** and **process each page immediately** after we get it. We simply **sum up the resources requested** by each pod (excluding completed pods) to get a set of resource quantities.
2. **Once we are done listing, we start a watch request** based on the resource version obtained from the list request above. Because pod resource requests cannot change, we only need to care about when a pod is created, completed, or deleted. **When a pod is created, we add its requested resources**, and **when a pod is completed/deleted, we subtract its requested resources** (the requested resources can be obtained from the object in the watch event). 
3. To avoid double counting when a completed pod gets deleted, we maintain the set of uids for completed pods that we have accounted for but are not yet deleted.
4. **When a watch connection drops, we freeze the current set of resource quantities, and start a (paginated) relist** where we add up the requested resources again from 0 to obtain a new set of resource quantities. Once the list is complete, we simply replace the old set with the new set and continue from step 2.

**Memory improvements**

(In relation to collecting total requested resources)

Previously, a copy of every pod in each member cluster was stored in memory. During relists, where we have to deserialize a complete list of pods again, the memory might even double.

With this new mechanism, the amount of memory used would be O(number of member clusters * page size). The amount of memory used for maintaining the set of completed-but-not-terminated-pods' uids should be insignificant.

**Results**

I ran two versions of the KubeAdmiral controller-manager against test clusters where I simulated the creation and deletion of 10000 pods at regular intervals. The first version is the main branch, while the second version uses the new cacheless collector. Periodic relists were triggered by pausing the apiservers to ensure that relists do not affect memory usage and correctness

I also hooked up some temporary metrics for evaluating the memory usage and correctness. Below are the results:

**Heap alloc of the 2 versions:**
image

**Reported CPU available (both correspond)**
image

**Reported memory available (both correspond)**
image


go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	github.com/evanphx/json-patch v5.6.0+incompatible
 	github.com/google/go-cmp v0.5.9
 	github.com/jinzhu/copier v0.3.5
-	github.com/onsi/ginkgo/v2 v2.8.2
-	github.com/onsi/gomega v1.27.0
+	github.com/onsi/ginkgo/v2 v2.9.1
```
SOF3, line 10: can we get dependabot instead

limhawjia: let's make an issue for that

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -93,8 +80,19 @@ func collectIndividualClusterStatus(
...
 
 	if !skip {
-		if err := updateClusterResources(ctx, &cluster.Status, clusterKubeInformer); err != nil {
-			return fmt.Errorf("failed to get cluster node levels: %w", err)
+		collector, err := c.getResourceCollectorForCluster(cluster.Name)
+		if err != nil {
+			return fmt.Errorf("failed to start resource collector: %w", err)
+		}
+
+		timeout, cancel := context.WithTimeout(context.TODO(), time.Second*5)
```
gary-lgy, line 88:
```suggestion
		timeout, cancel := context.WithTimeout(ctx, time.Second*5)
```

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -117,6 +115,47 @@ func collectIndividualClusterStatus(
...
+
+func (c *FederatedClusterController) deleteResourceCollectorForCluster(clusterName string) {
+	cached, ok := c.resourceCollectors.Load(clusterName)
+	if !ok {
+		return
+	}
+
+	collector := cached.(*clustercollector.ClusterCollector)
+	collector.Stop()
+	c.resourceCollectors.Delete(clusterName)
```
gary-lgy, line 156: Do we need to `LoadAndDelete` here? Any chance for a race condition between `getResourceCollectorForCluster` and `deleteResourceCollectorForCluster`?

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -93,8 +80,19 @@ func collectIndividualClusterStatus(
...
-		if err := updateClusterResources(ctx, &cluster.Status, clusterKubeInformer); err != nil {
-			return fmt.Errorf("failed to get cluster node levels: %w", err)
+		collector, err := c.getResourceCollectorForCluster(cluster.Name)
+		if err != nil {
+			return fmt.Errorf("failed to start resource collector: %w", err)
+		}
+
+		timeout, cancel := context.WithTimeout(context.TODO(), time.Second*5)
+		defer cancel()
+		if !cache.WaitForCacheSync(timeout.Done(), collector.HasSynced) {
```
gary-lgy, line 90: Can we use `WaitForNamedCacheSync` here

pkg/controllers/federatedcluster/controller.go
```diff
@@ -267,18 +265,19 @@ func (c *FederatedClusterController) collectClusterStatus(qualifiedName common.Q
 	}()
 
 	cluster, err := c.clusterLister.Get(qualifiedName.Name)
-	if err != nil && apierrors.IsNotFound(err) {
```
gary-lgy, line 270: Why do we log an error when the cluster is not found?

gary-lgy: Misread the diff location.

pkg/controllers/federatedcluster/clusterstatus.go
gary-lgy: Both methods are called from 2 different workers.

pkg/controllers/util/clustercollector/collector.go
```diff
@@ -0,0 +1,301 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
gary-lgy, line 2: Is the header correct?

pkg/controllers/util/federatedclient/client.go
```diff
@@ -97,8 +95,7 @@ func NewFederatedClientsetFactory(
 			newCluster := newObj.(*fedcorev1a1.FederatedCluster)
 
 			// only enqueue on generation change or cluster join condition changes
-			if oldCluster.Generation != newCluster.Generation ||
-				util.IsClusterJoined(&oldCluster.Status) != util.IsClusterJoined(&newCluster.Status) {
+			if util.IsClusterJoined(&oldCluster.Status) != util.IsClusterJoined(&newCluster.Status) {
```
gary-lgy, line 98: What's the reason behind this change?
Also the comment is outdated.

gary-lgy: Found the reason below.

test/e2e/framework/clusterprovider/kwokprovider.go
```diff
@@ -162,6 +162,7 @@ func (p *kwokClusterProvider) NewCluster(ctx context.Context, name string) (*fed
 func (p *kwokClusterProvider) StopCluster(ctx context.Context, name string) {
 	if _, ok := p.clusters.Load(name); ok {
```
gary-lgy, line 163: Do we need `LoadAndDelete` here?

test/e2e/federatedcluster/clusterstatus.go
```diff
@@ -124,18 +125,10 @@ var _ = ginkgo.Describe("Cluster Status", federatedClusterTestLabels, func() {
...
-				FederatedClusters().
-				Patch(ctx, cluster.Name, types.JSONPatchType, invalidClusterEndpointPatchData, metav1.PatchOptions{})
-			gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
-		}
-
 		ginkgo.Context("Without service account", func() {
 			ginkgo.It("Should succeed", func(ctx ginkgo.SpecContext) {
 				ginkgo.By("Creating cluster")
-				cluster, _ = f.NewCluster(ctx, framework.WithTaints)
+				cluster, _, stopClusterFn = f.NewCluster(ctx, framework.WithTaints)
```
gary-lgy, line 131: Why do we return a `stopClusterFn` instead of letting the caller use `StopCluster` directly?

gary-lgy (commented):
https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch Looks like we might not need to list

limhawjia:
> https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch Looks like we might not need to list

@gary-lgy This does not work because watch with rv=0 does not tell you when the added synthetic event stops (giving us a consistent snapshot). This means that we will not know when to "replace" the current resource state.

It would be great if it does though. It would remove all the overhead of listing and we would also not have to tinker with the rather confusing list semantics :(

pkg/controllers/federatedcluster/clusterstatus.go
limhawjia: Hmmm, I don't see a race condition. Could you be more specific?

pkg/controllers/federatedcluster/controller.go
limhawjia: Right... Good catch.

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+// ListWatcher lists and watches a given resource, emitting events to registered receivers. This allows receivers to
+// track changes to a resource starting from a consistent snapshot.
+//
+// Before any list request, ListWatcher must emit a StartSync event to notify receivers. This allows receivers to start
+// performing a resync if required. After all the objects from a successful list request is emitted, ListWatcher must
+// also emit a EndSync event to notify receivers.
+//
+// The design of ListWatcher and its implementations is loosely based on client-go's reflector.
```
gary-lgy, line 46: In `client-go` `ListerWatcher` is a lower-level component than `Reflector`, but `ListWatcher` is baed on `Reflector`. Would this be a source of confusion? Any chance we can name this `PaginatedReflector`?

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+	receivers []chan<- Event
+	pageSize  int64
+
+	mu      sync.Mutex
+	mux     chan Event
+	backoff wait.BackoffManager
+}
+
+// NewPagedListWatcher returns a ListWatcher implementation that uses paging when listing from the apiserver. Each page
+// is processed immediately to allow the allocated memory to be potentially garbage collected before the next request.
```
gary-lgy, line 86: Could we add a comment here to explain what we are optimising and at what cost.

test/e2e/federatedcluster/clusterstatus.go
limhawjia: I agree that `ListWatcher` is not the best name... But I feel `PaginatedReflector` is also not ideal, the name `Reflector` is based on the idea that a reflector "reflects" events into a store, which does happen here

test/e2e/framework/clusterprovider/kwokprovider.go
limhawjia: I think both works, but the reason why I returned a `stopClusterFn` is because I was not sure how to extend the interface of `Framework`, specifically what type of argument should be used to identify the cluster to be stopped - name, uid, cluster object?

_Preface:  `StopCluster` is currently only a method of the `ClusterProvider` interface, which is not directly exposed to users of the framework._

test/e2e/federatedcluster/clusterstatus.go
gary-lgy: > I think both works, but the reason why I returned a stopClusterFn is because I was not sure how to extend the interface of Framework, specifically what type of argument should be used to identify the cluster to be stopped - name, uid, cluster object?
>
> Preface: StopCluster is currently only a method of the ClusterProvider interface, which is not directly exposed to users of the framework.

You replied this on the wrong thread... For `stopCluster`, I think it's reasonable to expose it on the `Framework`. I think taking the cluster name should be sufficient. WDYT?

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+	logger := klog.FromContext(ctx).WithValues("component", "paged-list-watcher")
+	listCh := make(chan string, 1)
+
+	go func() {
+		for {
+			// list resource
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
gary-lgy, line 115: nit: we can move the main logic outside and select and eliminate some indentation

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+
+				p.mux <- Event{Type: EndSync}
+				listCh <- resourceVersion
+			}
+
+			// watch resource
+			select {
+			case <-ctx.Done():
+				return
+			case resourceVersion := <-listCh:
```
gary-lgy, line 133: Why do we need `listCh`?

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+				return "", fmt.Errorf("failed to list resource: %w", err)
+			}
+
+			listOpts.Continue = list.GetContinue()
+			resourceVersion = list.GetResourceVersion()
+
+			err = meta.EachListItem(resp, func(o runtime.Object) error {
+				p.mux <- Event{
+					Type:   Add,
+					Object: o.DeepCopyObject(),
```
gary-lgy, line 183: Is this DeepCopy necessary? This response will GC'd anyway right?

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+	listOpts := metav1.ListOptions{
+		Limit:    p.pageSize,
+		Continue: "",
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return "", nil
+		default:
```
gary-lgy, line 166: nit (same as above): can flatten this default branch

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+		if len(listOpts.Continue) == 0 {
+			break
+		}
+	}
+
+	return resourceVersion, nil
+}
+
+func (p *pagedListWatcher) watch(ctx context.Context, resourceVersion string) error {
+	watcher, err := p.lw.Watch(metav1.ListOptions{ResourceVersion: resourceVersion})
```
gary-lgy, line 202: https://github.com/kubernetes/client-go/blob/8cbca742aebe24b24f7f4e32fd999942fa9133e8/tools/cache/reflector.go#L315

Do we need to handle these errors?

```go
		w, err := r.listerWatcher.Watch(options)
		if err != nil {
			// If this is "connection refused" error, it means that most likely apiserver is not responsive.
			// It doesn't make sense to re-list all objects because most likely we will be able to restart
			// watch where we ended.
			// If that's the case begin exponentially backing off and resend watch request.
			// Do the same for "429" errors.
			if utilnet.IsConnectionRefused(err) || apierrors.IsTooManyRequests(err) {
				<-r.initConnBackoffManager.Backoff().C()
				continue
			}
			return err
		}
```

gary-lgy, line 202: https://github.com/kubernetes/client-go/blob/8cbca742aebe24b24f7f4e32fd999942fa9133e8/tools/cache/reflector.go#L302

Do we need `TimeoutSeconds` and `AllowWatchBookmarks`?

```go
		options := metav1.ListOptions{
			ResourceVersion: r.LastSyncResourceVersion(),
			// We want to avoid situations of hanging watchers. Stop any watchers that do not
			// receive any events within the timeout window.
			TimeoutSeconds: &timeoutSeconds,
			// To reduce load on kube-apiserver on watch restarts, you may enable watch bookmarks.
			// Reflector doesn't assume bookmarks are returned at all (if the server do not support
			// watch bookmarks, it will ignore this field).
			AllowWatchBookmarks: true,
		}
```

pkg/controllers/util/listwatcher/listwatcher.go
```diff
@@ -0,0 +1,239 @@
...
+				p.mux <- Event{
+					Type:   Update,
+					Object: event.Object,
+				}
+			case watch.Deleted:
+				p.mux <- Event{
+					Type:   Delete,
+					Object: event.Object,
+				}
+			case watch.Error:
```
gary-lgy, line 232: https://github.com/kubernetes/client-go/blob/8cbca742aebe24b24f7f4e32fd999942fa9133e8/tools/cache/reflector.go#L332

Do we need to handle these watch errors?
```go
			if err != errorStopRequested {
				switch {
				case isExpiredError(err):
					// Don't set LastSyncResourceVersionUnavailable - LIST call with ResourceVersion=RV already
					// has a semantic that it returns data at least as fresh as provided RV.
					// So first try to LIST with setting RV to resource version of last observed object.
					klog.V(4).Infof("%s: watch of %v closed with: %v", r.name, r.expectedTypeName, err)
				case apierrors.IsTooManyRequests(err):
					klog.V(2).Infof("%s: watch of %v returned 429 - backing off", r.name, r.expectedTypeName)
					<-r.initConnBackoffManager.Backoff().C()
					continue
				case apierrors.IsInternalError(err) && retry.ShouldRetry():
					klog.V(2).Infof("%s: retrying watch of %v internal error: %v", r.name, r.expectedTypeName, err)
					continue
				default:
					klog.Warningf("%s: watch of %v ended with: %v", r.name, r.expectedTypeName, err)
				}
			}
```

pkg/controllers/federatedcluster/clusterstatus.go
gary-lgy: > Both methods are called from 2 different workers.

I think I might have misread the code.

Will these 2 methods ever be called for the same cluster on 2 goroutines? If not, there won't be a race condition.

limhawjia:
1. Could you be more specific about what edge cases we need to handle? We could port the error-handling logic over as well.
2. We could simply not do a paginated list, but still only maintain a resource vector. This will still reduce the long-term memory usage but we will expect to see bursts during the initial list and relists.

limhawjia:
Nonetheless, we could close this pr if the risks are too large, it is just an optimization after all.

SOF3:
Alternative idea: Our concern was mainly that number of clusters (**m**) multiplied by number of pods (**n**) is too high. The memory required for the pods on a single cluster O(**n**) is expected to be bounded under an acceptable threshold, because that is exactly how other pod-watching components for a single cluster work (kube-scheduler, controller-manager, etc), so our actual requirement is just to keep the memory in O(**n**) instead of O(**mn**). We could more trivially solve this problem by using reflector to watch a single cluster, and using a semaphore to ensure that no more than **k** clusters are listed at the same time (and ensure the memory can be garbage collected before releasing the semaphore), then the memory complexity because O(**kn**) instead, which would be more acceptable.

gary-lgy:
> 1. Could you be more specific about what edge cases we need to handle? We could port the error-handling logic over as well.

I'm not sure which edge cases we need to handle, and that's the problem. Rather than porting `Reflector` to our own rewrite, I think it's would be more maintainable in the long run to just copy `Reflector`, make minimal changes and document exactly which lines are changed. An even better approach is to use `Reflector` and pass in custom `ListerWatcher` and `cache.Store` implementations that do what we want.

> 2. We could simply not do a paginated list, but still only maintain a resource vector. This will still reduce the long-term memory usage but we will expect to see bursts during the initial list and relists.

I think @SOF3's response hits the hail on the head. Reducing the memory usage to O(n) would be sufficient, and it can be done by using `Reflector` directly with a custom `ListerWatcher` and `cache.Store`.

PR #15, opened 2023-04-04
fix(override): manual overrides are not applied

gary-lgy:
- allow manual overrides to be applied
- reword override path description


limhawjia (approved):
lgtm

PR #16, opened 2023-04-04
test(e2e): implement generic resource propagation framework

gary-lgy:
Refactor the existing resource propagation tests using a generic (using golang generics) framework, allowing the easy addition of e2e tests for new types of resources while preventing code duplication.


limhawjia (approved):
This is great 🙌

PR #17, opened 2023-04-04
feat(auto-migrate): implement auto migration

gary-lgy:
This PR implements auto-migration, i.e. moving unschedulable pods to another cluster automatically.
To do so, the replica planner needs to be aware of the max number of pods that can be run in each cluster (`estimatedCapacity`) and plan accordingly. `estimatedCapacity` is calculated as `status.replicas - count(replicas unschedulable beyond a certain threshold`).

The KubeFed planner was `estimatedCapacity` aware, but had 2 issues:
- replica distribution will deviate significantly from preferences when `rebalance = false` and scaling down
- need to cache all pods from all clusters, which becomes a significant memory bottleneck

The later `refinedplanner` addressed these issues, but is unfortunately not aware of `estimatedCapacity`.

This PR is an attempt at creating a new replica distribution algorithm by combining the two planners, without having to cache all pods from all clusters. An `auto-migration` controller detects `estimatedCapacity` changes, and triggers a rescheduling. On rescheduling, the scheduler takes in the latest `estimatedCapcity` information and calculates the optimal replica distribution using the new algorithm.

This PR also fixes a few issues with the scheduler:
- sort taints correctly in scheduling trigger
- use a more idiomatic way to sort maps
- remove cluster ready conditions from scheduling trigger
- on cluster change, only enqueue objects if labels or taints have changed
- standardise logging levels
- fix incorrect handling of pending controllers resulting in spurious updates

This PR also improves infra:
- support post-generation CRD patches
- `AssertForItems` async matcher for e2e tests


pkg/controllers/federate/util.go
```diff
@@ -203,6 +203,7 @@ var (
 		common.FollowersAnnotation,
 	)
 
+	// TODO: Do we need to specify the internal annotations here?
```
limhawjia, line 206: are these the annotations that will be ignored when generating the template for the federated object?

if that is the case and we do not specify the internal annotations here, would it result in unnecessary updates to the federated object, and consequently unnecessary syncs to the member clusters when the internal annotations change?

pkg/controllers/util/planner/planner.go
```diff
@@ -103,15 +77,108 @@ func NewPlanner(preferences *ReplicaSchedulingPreference) *Planner {
...
-func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[string]int64,
-	estimatedCapacity map[string]int64, replicaSetKey string) (map[string]int64, map[string]int64, error) {
+func Plan(
+	rsp *ReplicaSchedulingPreference,
+	totalReplicas int64,
+	availableClusters []string,
+	currentReplicaCount map[string]int64,
+	estimatedCapacity map[string]int64,
+	replicaSetKey string,
+	avoidMigration bool,
```
limhawjia, line 87: Can we modify the name of this argument to `avoidDisruption` for consistency?

limhawjia:
Looks good other than the small issues that I raised. It would also be nice if we can formalize the algorithm, and have a doc to explain the feature to users.

pkg/controllers/federate/util.go
gary-lgy: No, as discussed offline, these are the annotations that should be ignored when computing template annotations from **source annotations**. For annotations that only appear on the federated object, there's no need to include them here. In fact the internal annotations listed here are redundant since they only appear on the federated object.

limhawjia (approved):

PR #19, opened 2023-04-04
feat(metrics): Prometheus collector

SOF3:
recreates #5


limhawjia:
is it possible to also export the default client-go metrics in this PR? i believe this can be done by importing the package `k8s.io/component-base/metrics/prometheus/clientgo`.

gary-lgy:
Closed in favor of #190 which is largely based on this PR.

PR #21, opened 2023-04-05
fix(rsp): current pod resource usage should be included in resource

SOF3:
calculation


pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -184,7 +184,7 @@ func updateClusterResources(
 		}
 	}
 
-	allocatable, available := aggregateResources(nodes, pods)
```
gary-lgy, line 187: Can we move this into utils so controller do not depend on each other?

SOF3:
I replaced the apiserver list request with a label-indexed pod cache in the pull request here: https://github.com/SOF3/kubeadmiral/pull/1

I wrote a few simple unit tests but I am not very confident if it really works as expected.

PR #27, opened 2023-04-13
feat: dynamic scheduling profiles

limhawjia:
This PR implements dynamic scheduling profiles for the scheduler.

Multiple scheduling profiles can be created dynamically using the newly updated `SchedulingProfile` CRD.

```yaml
apiVersion: core.kubeadmiral.io/v1alpha1
kind: SchedulingProfile
metadata:
    name: example-scheduling-profile
spec:
    plugins:
        filter:
            enabled:
            - name: XXX
            - name: XXX
            disabled:
            - name: XXX
        score:
            enabled:
            - type: Webhook
              name: XXX
              weight: 2
             disabled:
        select:
    pluginConfig:
    - name: my-rsp
       args:
           supplyLimitRatio: 1.4
```

Scheduling profiles are added to a propagation policy using the `profile` field.

```yaml
kind: PropagationPolicy
metadata:
    name: example-policy
spec:
    profile: example-scheduling-profile
```

To achieve this, `runtime.Framework`'s constructor now takes in an additional `EnabledPlugins` struct. This struct specifies the exact plugins to enable at which extension point by name.

```go
type EnabledPlugins struct {
	FilterPlugins   []string
	ScorePlugins    []string
	SelectPlugins   []string
	ReplicasPlugins []string
}

func NewFramework(registry Registry, handle framework.Handle, enabledPlugins *EnabledPlugins) (framework.Framework, error) {
...
}
```

In the reconcile loop of the scheduler, the `getProfileForFederatedObject` method call will now translate any attached `SchedulingProfile` into an `EnabledPlugins` struct, resolving it against the default set of plugins, before passing it to the constructor for `runtime.Framework`.

```go
func (s *Scheduler) profileForFedObject(
	_ *unstructured.Unstructured,
	profile *fedcorev1a1.SchedulingProfile,
	handle framework.Handle,
) (framework.Framework, error) {
	enabledPlugins := getDefaultEnabledPlugins()
	if profile != nil {
		applyProfile(enabledPlugins, profile)
	}

	// TODO: merge the inTree registry with a dynamically generated webhook registry when supporting webhook plugins

	return runtime.NewFramework(
		inTreeRegistry,
		handle,
		enabledPlugins,
	)
}
```

The returned `runtime.Framework` will only contain plugins that have been enabled. This allows users to customize the plugins to run during the scheduling of a resource.

**NOTE**

This PR does implement the respecting of `Weight` specified for score plugins. This will be done in a later PR when we are more sure about the approach to take with regard to cluster score normalization.


pkg/apis/core/v1alpha1/types_schedulingprofile.go
```diff
@@ -26,7 +26,6 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:validation:Required
 // +kubebuilder:resource:path=schedulingprofiles,shortName=sp,singular=schedulingprofile,scope=Cluster
-// +kubebuilder:object:root=true
```
gary-lgy, line 29: Why did we remove this?

gary-lgy: Looks like `+k8s:deepcopy-gen` and `+k8s:deepcopy-gen:interface` have been deprecated in favor of `+kubebuilder:object:generate` and `+kubebuilder:object:root`.  https://book.kubebuilder.io/reference/markers/object.html

pkg/apis/core/v1alpha1/types_schedulingprofile.go
```diff
@@ -46,15 +45,74 @@ type SchedulingProfileList struct {
...
 type SchedulingProfileSpec struct {
-	// Plugins is the list of plugins to use when scheduling a resource
-	Plugins []SchedulingPlugin `json:"plugins"`
+	// Plugins specify the set of plugins that should be enabled or disabled.
+	// Enabled plugins are the ones that should be enabled in addition to the
+	// default plugins. Disabled plugins are any of the default plugins that
+	// should be disabled.
+	// When no enabled or disabled plugin is specified for an extension point,
+	// default plugins for that extension point will be used if there is any.
+	Plugins *Plugins `json:"plugins"`
```
gary-lgy, line 54: Should this field be `omitempty` and `+optional`? In case we only want to customise the plugin config

pkg/apis/core/v1alpha1/types_schedulingprofile.go
```diff
@@ -46,15 +45,74 @@ type SchedulingProfileList struct {
...
+	// be invoked before default plugins, default plugins must be disabled and re-enabled here in desired order.
+	// +optional
+	Enabled []Plugin `json:"enabled,omitempty"`
+	// Disabled specifies default plugins that should be disabled.
+	// +optional
+	Disabled []Plugin `json:"disabled,omitempty"`
+}
+
+// Plugin specifies a plugin type, name and its weight when applicable. Weight is used only for Score plugins. Type
+// should be ommited when referencing in-tree plugins.
```
gary-lgy, line 90:
```suggestion
// should be omitted when referencing in-tree plugins.
```

pkg/apis/core/v1alpha1/types_schedulingprofile.go
```diff
@@ -46,15 +45,74 @@ type SchedulingProfileList struct {
...
+type Plugin struct {
+	// Type defines the type of the plugin.
+	// +kubebuilder:validation:Enum=Webhook
+	// +optional
+	Type PluginType `json:"type,omitempty"`
+	// Name defines the name of the plugin.
+	Name string `json:"name,omitempty"`
+	// Weight defines the weight of the plugin.
+	// +optional
+	Weight int64 `json:"wait,omitempty"`
```
gary-lgy, line 100: Can we add a validation for >0 here

pkg/apis/core/v1alpha1/types_schedulingprofile.go
```diff
@@ -46,15 +45,74 @@ type SchedulingProfileList struct {
...
+	// Disabled specifies default plugins that should be disabled.
+	// +optional
+	Disabled []Plugin `json:"disabled,omitempty"`
+}
+
+// Plugin specifies a plugin type, name and its weight when applicable. Weight is used only for Score plugins. Type
+// should be ommited when referencing in-tree plugins.
+type Plugin struct {
+	// Type defines the type of the plugin.
+	// +kubebuilder:validation:Enum=Webhook
```
gary-lgy, line 93: Is it better to specify this on the type instead?

pkg/controllers/scheduler/framework/runtime/framework.go
```diff
@@ -41,53 +67,72 @@ type frameworkImpl struct {
...
 var _ framework.Framework = &frameworkImpl{}
 
-func NewFramework(registry Registry, handle framework.Handle) (framework.Framework, error) {
+func NewFramework(registry Registry, handle framework.Handle, enabledPlugins *EnabledPlugins) (framework.Framework, error) {
 	fwk := &frameworkImpl{}
 
 	pluginsMap := make(map[string]framework.Plugin)
 
 	for name, factory := range registry {
+	if !enabledPlugins.isPluginEnabled(name) {
```
gary-lgy, line 76: wrong indentation?

pkg/controllers/scheduler/framework/runtime/framework.go
```diff
@@ -41,53 +67,72 @@ type frameworkImpl struct {
...
-			continue
+	registeredPlugins := sets.New[string]()
+	for _, plugin := range enabledPlugins {
+		pg, ok := pluginsMap[plugin]
+		if !ok {
+			return fmt.Errorf("%s %s does not exist", pluginType.Name(), plugin)
+		}
+
+		if !reflect.TypeOf(pg).Implements(pluginType) {
+			return fmt.Errorf("plugin %s does not extend %s", plugin, pluginType.Name())
```
gary-lgy, line 107:
```suggestion
			return fmt.Errorf("plugin %s does not implement %s", plugin, pluginType.Name())
```

pkg/controllers/scheduler/framework/runtime/framework.go
```diff
@@ -41,53 +67,72 @@ type frameworkImpl struct {
...
+func addPlugins(pluginList interface{}, enabledPlugins []string, pluginsMap map[string]framework.Plugin) error {
 	plugins := reflect.ValueOf(pluginList).Elem()
 	pluginType := plugins.Type().Elem()
 
-	for _, plugin := range pluginsMap {
-		if !reflect.TypeOf(plugin).Implements(pluginType) {
-			// klog.Infof("plugin %q does not extend %s plugin", name, pluginType.Name())
-			continue
+	registeredPlugins := sets.New[string]()
+	for _, plugin := range enabledPlugins {
```
gary-lgy, line 100: codestyle nit: 
```suggestion
	for _, pluginName := range enabledPluginNames {
		plugin, ok := pluginsMap[pluginName]
```

pkg/controllers/scheduler/framework/runtime/framework.go
```diff
@@ -41,53 +67,72 @@ type frameworkImpl struct {
...
+		pg, ok := pluginsMap[plugin]
```
gary-lgy, line 101: codestyle nit: 
```suggestion
	for _, pluginName := range enabledPluginNames {
		plugin, ok := pluginsMap[pluginName]
```

pkg/controllers/scheduler/framework/runtime/framework_test.go
```diff
@@ -113,3 +203,112 @@ func TestRunFilterPlugins(t *testing.T) {
...
+			true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			fwk, err := NewFramework(registry, nil, &test.enabledPlugins)
+			if test.shouldError {
+				if err == nil {
+					t.Errorf("expected error when creating framework but got nil")
```
gary-lgy, line 299:
```suggestion
					t.Fatalf("expected error when creating framework but got nil")
```
`Errorf` does not terminate the current test.

pkg/controllers/scheduler/framework/runtime/framework_test.go
```diff
@@ -113,3 +203,112 @@ func TestRunFilterPlugins(t *testing.T) {
...
+				}
+				t.Logf("got expected error: %v", err)
+				return
+			}
+
+			if err != nil {
+				t.Errorf("unexpected error when creating framework: %v", err)
```
gary-lgy, line 306:
```suggestion
				t.Fatalf("unexpected error when creating framework: %v", err)
```

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: it was a mistake

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: yes i think so, overlooked this part

limhawjia:
@gary-lgy fixed the problems, along with a few more lint errors

pkg/controllers/scheduler/framework/runtime/framework_test.go
```diff
@@ -113,3 +203,112 @@ func TestRunFilterPlugins(t *testing.T) {
...
+				ScorePlugins:    []string{"score", "scoreAndSelect"},
+				SelectPlugins:   []string{"scoreAndSelect", "select", "notexists"},
+				ReplicasPlugins: []string{"replicas"},
+			},
+			nil,
+			true,
+		},
+	}
+
+	for _, test := range tests {
```
gary-lgy, line 294: Do we need to test that:
- disabled plugins will not be instantiated
- enabled plugins will only be instantiated once?

pkg/controllers/scheduler/profile.go
```diff
@@ -30,19 +32,94 @@ import (
...
-		clusterresources.ClusterResourcesLeastAllocatedName:     clusterresources.NewClusterResourcesLeastAllocated,
-		rsp.ClusterCapacityWeightName:                           rsp.NewClusterCapacityWeight,
-		placement.PlacementFilterName:                           placement.NewPlacementFilter,
-		clusteraffinity.ClusterAffinityName:                     clusteraffinity.NewClusterAffinity,
-		maxcluster.MaxClusterName:                               maxcluster.NewMaxCluster,
-	}
-
-	return runtime.NewFramework(DefaultRegistry, handle)
+// inTreeRegistry should contain all known in-tree plugins
+var inTreeRegistry = runtime.Registry{
```
gary-lgy, line 36: Should this be versioned?

pkg/controllers/scheduler/profile.go
```diff
@@ -30,19 +32,94 @@ import (
...
+	apiresources.APIResourcesName:                           apiresources.NewAPIResources,
+	clusteraffinity.ClusterAffinityName:                     clusteraffinity.NewClusterAffinity,
+	clusterresources.ClusterResourcesFitName:                clusterresources.NewClusterResourcesFit,
+	placement.PlacementFilterName:                           placement.NewPlacementFilter,
+	tainttoleration.TaintTolerationName:                     tainttoleration.NewTaintToleration,
+	clusterresources.ClusterResourcesBalancedAllocationName: clusterresources.NewClusterResourcesBalancedAllocation,
+	clusterresources.ClusterResourcesLeastAllocatedName:     clusterresources.NewClusterResourcesLeastAllocated,
+	clusterresources.ClusterResourcesMostAllocatedName:      clusterresources.NewClusterResourcesMostAllocated,
```
gary-lgy, line 44: Why are we enabling Most and LeastAllocated at the same time?

gary-lgy: nvm, realised this is the full list and not the enabled list.

pkg/controllers/scheduler/profile.go
```diff
@@ -30,19 +32,94 @@ import (
...
+	maxcluster.MaxClusterName:                               maxcluster.NewMaxCluster,
+	rsp.ClusterCapacityWeightName:                           rsp.NewClusterCapacityWeight,
```
gary-lgy, line 46: Did we intentionally omit `Fit`?

pkg/controllers/scheduler/profile.go
limhawjia: This is just the registry, not the enabled plugins

pkg/controllers/scheduler/profile.go
limhawjia: Nope, this is a mistake. Luckily you caught it

pkg/controllers/scheduler/profile.go
```diff
@@ -30,19 +32,94 @@ import (
...
+	}
+
+	for _, p := range pluginSet.Enabled {
+		result = append(result, p.Name)
+	}
+
+	return result
+}
+
+func (s *Scheduler) profileForFedObject(
```
gary-lgy, line 108:
```suggestion
func (s *Scheduler) createFramework(
```

pkg/controllers/scheduler/profile.go
limhawjia: Hmm... that's an interesting suggestion, but i don't see the need to... In any case, there is no versioning of in-tree plugins exposed to the users anyway.

pkg/controllers/scheduler/profile.go
limhawjia: Wait, it is included already, you might have missed it above

pkg/controllers/scheduler/profile_test.go
```diff
@@ -0,0 +1,650 @@
...
+					"expected replicas plugins to be %v, but got %v",
+					test.base.ReplicasPlugins,
+					test.expectedResult.ReplicasPlugins,
+				)
+			}
+		})
+	}
+}
+
+func checkPluginsEqual(result []string, expected []string) bool {
```
gary-lgy, line 639: Why not `DeepEqual`?

pkg/controllers/scheduler/scheduler.go
```diff
@@ -339,16 +341,41 @@ func (s *Scheduler) reconcile(qualifiedName common.QualifiedName) (status worker
...
+				s.eventRecorder.Eventf(
+					fedObject,
+					corev1.EventTypeWarning,
+					EventReasonScheduleFederatedObject,
+					"failed to schedule object: %v",
+					fmt.Errorf("failed to get scheduling profile: %w", err),
+				)
+
+				if apierrors.IsNotFound(err) {
+					return worker.StatusErrorNoRetry
```
gary-lgy, line 361: From the controller's perspective, this is not an error - the user specified a non-existent profile and we didn't find it. The controller did everything it can do. Shouldn't we return `StatusAllOK` here?

pkg/controllers/scheduler/scheduler.go
```diff
@@ -339,16 +341,41 @@ func (s *Scheduler) reconcile(qualifiedName common.QualifiedName) (status worker
...
 		}
-		profile, err := s.profileForFedObject(fedObject, s.buildFrameworkHandle())
+
+		var schedulingProfile *fedcorev1a1.SchedulingProfile
+		profileName := policy.GetSpec().SchedulingProfile
+		if len(profileName) > 0 {
+			schedulingProfile, err = s.schedulingProfileLister.Get(profileName)
+			if err != nil {
+				// TODO(hawjia): handle error - should we retry if not found?
+				keyedLogger.Error(err, "Failed to get scheduling profile")
```
gary-lgy, line 351:
```suggestion
				keyedLogger.Error(err, "Failed to get scheduling profile", "profile", profileName)
```

pkg/controllers/scheduler/scheduler.go
```diff
@@ -339,16 +341,41 @@ func (s *Scheduler) reconcile(qualifiedName common.QualifiedName) (status worker
...
 			return worker.StatusError
 		}
-		profile, err := s.profileForFedObject(fedObject, s.buildFrameworkHandle())
+
+		var schedulingProfile *fedcorev1a1.SchedulingProfile
+		profileName := policy.GetSpec().SchedulingProfile
+		if len(profileName) > 0 {
+			schedulingProfile, err = s.schedulingProfileLister.Get(profileName)
+			if err != nil {
+				// TODO(hawjia): handle error - should we retry if not found?
```
gary-lgy, line 350: Is this TODO still valid?

pkg/controllers/scheduler/scheduler.go
```diff
@@ -339,16 +341,41 @@ func (s *Scheduler) reconcile(qualifiedName common.QualifiedName) (status worker
...
+				keyedLogger.Error(err, "Failed to get scheduling profile")
+				s.eventRecorder.Eventf(
+					fedObject,
+					corev1.EventTypeWarning,
+					EventReasonScheduleFederatedObject,
+					"failed to schedule object: %v",
+					fmt.Errorf("failed to get scheduling profile: %w", err),
```
gary-lgy, line 357: Can we include the profile name here

pkg/controllers/scheduler/schedulingtriggers.go
```diff
@@ -276,3 +276,55 @@ func (s *Scheduler) enqueueFederatedObjectsForCluster(cluster pkgruntime.Object)
 		s.worker.EnqueueObject(fedObject)
 	}
 }
+
+// enqueueFederatedObjectsForProfile enqueues all federated objects that match the profile
+func (s *Scheduler) enqueueFederatedObjectsForProfile(profile pkgruntime.Object) {
+	profileObj := profile.(*fedcorev1a1.SchedulingProfile)
+
+	s.logger.WithValues("profile", profileObj.GetName()).V(2).Info("Enqueue federated objects for profile")
```
gary-lgy, line 284: We should make a new `logger` instance here with the new KV embedded

pkg/controllers/scheduler/schedulingtriggers.go
```diff
@@ -276,3 +276,55 @@ func (s *Scheduler) enqueueFederatedObjectsForCluster(cluster pkgruntime.Object)
...
+
+	if s.typeConfig.GetNamespaced() {
+		policies, err := s.propagationPolicyLister.List(labels.Everything())
+		if err != nil {
+			s.logger.Error(err, "Failed to enqueue federated object for profile")
+			return
+		}
+		for _, policy := range policies {
+			if policy.Spec.SchedulingProfile == profileObj.GetName() {
+				qualifedName := common.NewQualifiedName(policy).String()
```
gary-lgy, line 296:
```suggestion
				qualified := common.NewQualifiedName(policy).String()
```

pkg/controllers/scheduler/schedulingtriggers.go
```diff
@@ -276,3 +276,55 @@ func (s *Scheduler) enqueueFederatedObjectsForCluster(cluster pkgruntime.Object)
 		s.worker.EnqueueObject(fedObject)
 	}
 }
+
+// enqueueFederatedObjectsForProfile enqueues all federated objects that match the profile
+func (s *Scheduler) enqueueFederatedObjectsForProfile(profile pkgruntime.Object) {
```
gary-lgy, line 281: Not a request for changes, just a general comment:
This looks very dangerous, do does `enqueueFederatedObjectsForPolicy`. We may need to rethink how we response to policy and profile events.

pkg/controllers/scheduler/schedulingtriggers.go
```diff
@@ -276,3 +276,55 @@ func (s *Scheduler) enqueueFederatedObjectsForCluster(cluster pkgruntime.Object)
 		s.worker.EnqueueObject(fedObject)
 	}
 }
+
+// enqueueFederatedObjectsForProfile enqueues all federated objects that match the profile
```
gary-lgy, line 280: Why do we enqueue the objects if the trigger hash stays the same? Is it only to handle add events? If that's the case, can we only respond to add events?

pkg/controllers/scheduler/scheduler_test.go
```diff
@@ -16,361 +16,361 @@ limitations under the License.
 
 package scheduler
 
-import (
-	"reflect"
-	"testing"
+// import (
```
gary-lgy, line 19: Why is this file commented out?

pkg/controllers/scheduler/scheduler.go
limhawjia: Agreed.

pkg/controllers/scheduler/schedulingtriggers.go
limhawjia: Yes, to handle add events. Since we have not decided how to respond to profile changes, we can ignore update and delete events for now

pkg/controllers/scheduler/scheduler.go
limhawjia: I'm gonna update `keyedLogger` at the start of this block with `["profile", profileName]` kv

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/kubewharf/kubeadmiral/test/e2e/framework/resources"
+)
+
+var scheduleTimeout = time.Second * 10
+
+var _ = ginkgo.Describe("Scheduling Profile", func() {
+	f := framework.NewFramework("scheduling-profile", framework.FrameworkOptions{CreateNamespace: true})
+
+	testPlacementFilter := func(ctx context.Context, profile *fedcorev1a1.SchedulingProfile, enabled bool) {
+		clusters, err := f.HostFedClient().CoreV1alpha1().FederatedClusters().List(ctx, metav1.ListOptions{})
```
gary-lgy, line 45: Do we need to filter out e2e test objects?

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,196 @@
...
+		if len(clusters.Items) == 0 {
+			ginkgo.Fail("Test requires at least one federated cluster")
+		}
+
+		ginkgo.By("Creating scheduling profile")
+		profile, err = f.HostFedClient().CoreV1alpha1().SchedulingProfiles().Create(ctx, profile, metav1.CreateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+		ginkgo.DeferCleanup(func(ctx ginkgo.SpecContext) {
+			if err := f.HostFedClient().CoreV1alpha1().SchedulingProfiles().Delete(ctx, profile.Name, metav1.DeleteOptions{}); err != nil &&
+				!apierrors.IsNotFound(err) {
```
gary-lgy, line 56: Why not assert error is NotFound or nil? A log may be less obvious

limhawjia: Will an assertion failure here cause the process to exit immediately and affect the rest of the cleanup steps?

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,196 @@
...
+				g.Expect(err).ToNot(gomega.HaveOccurred())
+
+				placementObj, err := util.UnmarshalGenericPlacements(federatedConfigMap)
+				gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+
+				placement := placementObj.Spec.GetPlacementOrNil(scheduler.PrefixedGlobalSchedulerName)
+				g.Expect(placement).ToNot(gomega.BeNil())
+
+				// only the first cluster should be selected since placement plugin is enabled
+				clusters, err := f.HostFedClient().CoreV1alpha1().FederatedClusters().List(ctx, metav1.ListOptions{})
```
gary-lgy, line 104: Do we need to list again here?

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,196 @@
...
+					"failed to cleanup scheduling profile",
+					"profile-name",
+					profile.Name,
+				)
+			}
+		})
+
+		ginkgo.By("Creating propagation policy that references scheduling profile")
+		policy := policies.PropagationPolicyForClustersWithPlacements(f.Name(), []*fedcorev1a1.FederatedCluster{
+			{
```
gary-lgy, line 68: Why create a new FC on the fly? Can't we use `clusters` on line 45?

pkg/controllers/scheduler/schedulingtriggers.go
limhawjia: Actually would it be better to remove all event handlers for profile until we are sure how to respond to it? cc @SOF3

pkg/controllers/scheduler/scheduler_test.go
limhawjia: reverted

test/e2e/schedulingprofile/intree.go
limhawjia: yes

test/e2e/schedulingprofile/intree.go
limhawjia: oops, yes i should've just used `clusters`

test/e2e/schedulingprofile/intree.go
limhawjia: no need

pkg/controllers/scheduler/schedulingtriggers.go
limhawjia: we can create an issue to discuss this

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,196 @@
...
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+
+		ginkgo.By("Creating configmap that references propagation policy")
+		configMap := resources.GetSimpleConfigMap(f.Name())
+		policies.SetPropagationPolicy(configMap, policy)
+		configMap, err = f.HostKubeClient().CoreV1().ConfigMaps(f.TestNamespace().Name).Create(ctx, configMap, metav1.CreateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+
+		ginkgo.By("Verifying scheduling result")
+		if enabled {
```
gary-lgy, line 90: nit: this if branch can be nested in the eventually since the logic for getting placement is shared

pkg/controllers/scheduler/framework/runtime/framework.go
limhawjia: fixed

pkg/controllers/scheduler/framework/runtime/framework_test.go
limhawjia: fixed

pkg/controllers/scheduler/framework/runtime/framework_test.go
limhawjia: fixed

pkg/controllers/scheduler/profile.go
limhawjia: fixed

pkg/controllers/scheduler/scheduler.go
limhawjia: fixed

pkg/controllers/scheduler/scheduler.go
limhawjia: fixed. removed

test/e2e/schedulingprofile/intree.go
limhawjia: fixed

test/e2e/schedulingprofile/intree.go
limhawjia: fixed

test/e2e/schedulingprofile/intree.go
limhawjia: fixed

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: fixed

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: added

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: fixed

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: fixed

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: fixed

pkg/apis/core/v1alpha1/types_schedulingprofile.go
gary-lgy: did you push?

pkg/controllers/scheduler/profile.go
gary-lgy: What about the enabledByDefault plugins?

test/e2e/schedulingprofile/intree.go
gary-lgy: From my understanding the clean up steps from container nodes should still be run, just like normal Subject nodes.

pkg/controllers/scheduler/schedulingtriggers.go
limhawjia: i'm going to remove the event handlers for scheduling profile until we get a better idea of how we want to deal with profile events

pkg/controllers/scheduler/framework/runtime/framework_test.go
gary-lgy: Comments for `PluginConfig` states that 
>  A plugin that is invoked at multiple extension points is initialized once.

These 2 tests are to verify this behavior.
1. If a profile disables a plugin that is present in the registry, we need to ensure its constructor is never called.
2. If a profile enables a plugin at multiple extension points, we need to ensure its constructor is called exactly once.

pkg/controllers/util/worker/constants.go
```diff
@@ -62,4 +62,7 @@ var (
 	// StatusConflict indicates that reconciliation could not be completed due to a conflict
 	// (e.g., AlreadyExists or Conflict), and the object should be requeued with a small delay.
 	StatusConflict = Result{Success: false, RequeueAfter: pointer.Duration(500 * time.Millisecond), Backoff: false}
+	// StatusErrorNoRetry indicates that the reconciliation could not be completed due to an error, but the object
```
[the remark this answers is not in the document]
gary-lgy: Can we delete this new status now that it's now used anymore

pkg/controllers/scheduler/schedulingtriggers.go
gary-lgy: Can we delete this function as well since it would be dead code

pkg/controllers/scheduler/profile.go
limhawjia: there's no versioning of that either... i think if we want to version the in-tree plugins or enabledbydefault plugins we can open a new issue to discuss it

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: yes

gary-lgy: I still see the marker comment on the field 
image
and this comment is not marked as outdated

pkg/controllers/scheduler/profile.go
gary-lgy: Didn't we discuss this already - the enabled+disabled way of constructing profiles depends on the default set of enabled plugins being fixed for a specific version - otherwise with the same profile you'll get different resolved plugins - breaking BC.

pkg/controllers/scheduler/schedulingtriggers.go
limhawjia: done

pkg/controllers/scheduler/framework/runtime/framework_test.go
gary-lgy: 1. `NewFramework` takes in `egistry` and `enabledPlugins`. If a plugin is in `registry` but not in `enabledPlugins`, we should make sure its constructor is never called.  We can do this by registering a plugin that panics on construction, for example. The tests in `profile_test.go` only test the plugin resolution, not the plugin construction.
2. We can also test this using a plugin that panics on repeated construction, for example.

limhawjia: got it.

1.  i will add a plugin that panics on construction
2. i guess all plugins should panic on repeated construction?

test/e2e/schedulingprofile/intree.go
```diff
@@ -0,0 +1,188 @@
...
+				gomega.Expect(err).To(gomega.Or(gomega.BeNil(), gomega.Satisfy(apierrors.IsNotFound)))
+				g.Expect(err).ToNot(gomega.HaveOccurred())
+
+				placementObj, err := controllerutil.UnmarshalGenericPlacements(federatedConfigMap)
+				gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+
+				placement := placementObj.Spec.GetPlacementOrNil(scheduler.PrefixedGlobalSchedulerName)
+				g.Expect(placement).ToNot(gomega.BeNil())
+
+				// all clusters should be selected since placement plugin was disabled
```
gary-lgy, line 119: If there's only 1 cluster, will both `enabled` and `!enabled` pass regardless of whether the placement filter is enabled?

pkg/controllers/scheduler/framework/runtime/framework_test.go
limhawjia: added some checks for 1 and 2

pkg/apis/core/v1alpha1/types_schedulingprofile.go
limhawjia: fixed now

pkg/controllers/scheduler/framework/runtime/framework_test.go
```diff
@@ -212,13 +213,56 @@ func TestNewFramework(t *testing.T) {
...
-		"select":         func(f framework.Handle) (framework.Plugin, error) { return selectPlugin, nil },
-		"replicas":       func(f framework.Handle) (framework.Plugin, error) { return replicasPlugin, nil },
-		"filterAndScore": func(f framework.Handle) (framework.Plugin, error) { return filterAndScorePlugin, nil },
-		"scoreAndSelect": func(f framework.Handle) (framework.Plugin, error) { return scoreAndSelectPlugin, nil },
+	newRegistry := func() Registry {
+		var filterConstructed, scoreConstructed, selectConstructed, replicasConstructed, filterAndScoreConstructed, scoreAndSelectConstructed atomic.Bool
+		return Registry{
+			"filter": func(f framework.Handle) (framework.Plugin, error) {
+				if filterConstructed.Load() {
+					t.Errorf("filter plugin constructed more than once")
```
gary-lgy, line 221: Can we use `t.Fatalf` here

pkg/controllers/scheduler/profile.go
limhawjia: versioning for default plugins is done 😁, as per our discussion, the in-tree registry does not need to be versioned

test/e2e/schedulingprofile/intree.go
limhawjia: right, the minimum should be 2. fixed

gary-lgy (approved):

PR #28, opened 2023-04-14
fix: limit pp/cpp length to 63

SOF3:
Label values cannot exceed 63, so pp/cpp names that are too long cannot work anyway


config/crds/patches/core.kubeadmiral.io_propagationpolicies.sh
```diff
@@ -0,0 +1,3 @@
+# controller-gen does not respect {} as default value for a struct field
+# issue: https://github.com/kubernetes-sigs/controller-tools/issues/622
```
gary-lgy, line 2: Can we move this comment (and the identical comment in CPP patch) into the generic patch?

gary-lgy (approved):

PR #29, opened 2023-04-14
feat(scheduler): webhook plugins

gary-lgy:
Previously scheduler plugins had to be written in-tree. This means adding a new plugin requires recompiling the controller-manager binary. This is tedious for developers and might be impossible for end users. This PR allows plugins to be implemented as remote HTTP webhooks, which can be used just like normal scheduler plugins.

This is especially useful for plugins that:
- schedule based on special business logic that cannot be generalized
- schedule based on external and/or existing systems that cannot be integrated into kubeadmiral
- are developed and deployed separately from the kubeadmiral core control plane

The webhook plugin design is heavily inspired by `kube-scheduler`'s [scheduler extenders](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/1819-scheduler-extender), with the difference that webhook plugins in kubeadmiral are treated as first-class plugins that can be enabled, disabled, or reordered just like native plugins.

A webhook can implement `Filter`, `Score`, and `Select` stages in the scheduling framework, and exposes these as HTTP endpoints expecting JSON payloads. The scheduler calls the webhook during scheduling cycles, passing in the arguments required for each stage, e.g.,`SchedulingUnit` and `FederatedCluster`s. In turn, the webhook should return the scheduling results encoded in JSON.

The webhook API is versioned separately from the core API version. A webhook can expect multiple payload versions and the scheduler chooses the first version that it supports. This enables smooth upgrade of the payload version without disruption.

A webhook can be configured using a `SchedulingPluginWebhookConfiguration` CRD.


limhawjia (approved):
lgtm

PR #30, opened 2023-04-15
fix(cluster): add ready condition when api discovery fails

gary-lgy:
When api discovery for a member cluster fails, the cluster controller errors out and does not set the ready and offline conditions on the `FederatedCluster`. This PR updates the status update logic to also set ready and offline conditions (with new reasons explaining the issue) when api discovery or resource collection for a cluster fails.

This PR also fixes a regression where cluster join timeout defaults to 0 seconds, and makes cluster condition matching in e2e tests more stringent.


limhawjia (approved):
lgtm

PR #31, opened 2023-04-17
ci: github actions for testing and linting

gary-lgy:
Example CI run for push to main: https://github.com/gary-lgy/kubeadmiral/actions/runs/4718894465
Example CI run for PR: https://github.com/gary-lgy/kubeadmiral/actions/runs/4718489828

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,66 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+
+# TODO: coverage
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version:
+          - "1.19"
+          - "1.20"
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - run: make test
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.19"
+          # https://github.com/golangci/golangci-lint-action/issues/244
+          # without `cache: false`, the action might intermittently fail
+          cache: false
+
+      - uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.52.2
+          only-new-issues: true
+          args: >
+            --timeout=5m
+            --max-issues-per-linter=0
+            --max-same-issues=0
+            --uniq-by-line=false
+
+  verify-codegen:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.19"
+
+      - run: |
+          make generate -B
+          git diff --exit-code && exit 0
+          echo "Codegen is not up to date. Please run 'make generate' and commit the changes." >&2
+          exit 1


.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+#           only-new-issues: true
+
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name: Go
+
+on:
+  push:
+    branches: ["main"]
```
SOF3, line 64: Why do we limit branches to `main`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+
+      - name: Build
+        run: go build -v ./...
+
+      - name: Test
+        run: go test -v ./...
```
SOF3, line 83: Should we `-race` here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+
+      - name: Build
+        run: go build -v ./...
```
SOF3, line 80: Is building necessary here? Wouldn't `go test` already reveal possible compile errors?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
+# name: CI
```
SOF3, line 1: Why two workflows here?

SOF3 (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+# TODO: do we need this?
+# permissions:
+#   contents: read # for actions/checkout to fetch code
```
SOF3, line 9: afaik this is pretty much implied. most workflows checkout without specifying this anyway.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Build
+        run: make manager
```
SOF3, line 31: Why do we need to build manager here?

gary-lgy: `go test` only builds packages that have unit tests, iirc.

gary-lgy: Looks like I'm wrong. From `go help test`:
> 'Go test' recompiles each package along with any files with names matching
the file pattern "*_test.go".

.github/workflows/ci.yml
gary-lgy: This is to make sure CI doesn't run twice for each push to an open PR

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.19"
+
+      - run: |
+          make generate -B
+          git diff --exit-code && exit 0
```
SOF3, line 64: If I remember correctly, github runs commands with `-e` enabled, so this line would just fail directly if there is diff. But it should be obvious enough anyway.

SOF3 (approved):

PR #46, opened 2023-04-27
feat: customize pod informer to reduce memory usage

JackZxj:
Use a custom Pod Informer to limit the number of concurrent list pods from member clusters, and simplify the metadata of the pods before we store them into the cache.

The reference conversations: https://github.com/kubewharf/kubeadmiral/pull/14

As result:
image
4 clusters, 100 kwok-node/cluster, 10 deployment/cluster, 1000 pod/deployment, total: 10k pod/cluster

* case1: simplify the pod (In the last three rounds, kwok failed to delete pods of a cluster so the pods of cluster came to 20k; and some kwok nodes got not ready, so the available cpu,gpu came to zero, but the pods still in work. The overall impact is not significant.)
* case2: regular pod informer
* case3: simplify the pod and limit 2 available listers 
* case4: simplify the pod, ignore modified event of watch, and limit 2 available listers

summary: It can save at least half of the memory, and reduce the rate of change of memory usage

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -53,6 +53,9 @@ type Options struct {
 	NSAutoPropExcludeRegexp string
 	CreateCRDsForFTCs       bool
 	ClusterJoinTimeout      time.Duration
+
+	MaxPodListers    int64
+	EnablePodPruning bool
 }
 
 func NewOptions() *Options {
@@ -109,6 +112,10 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		"The maximum amount of time to wait for a new cluster to join the federation before timing out.",
 	)
 
+	flags.Int64Var(&o.MaxPodListers, "max-pod-listers", 0, "The maximum number of concurrent pod listing requests to member clusters. "+
+		"A non-positive number means unlimited, but may increase the instantaneous memory usage.")
+	flags.BoolVar(&o.EnablePodPruning, "enable-pod-pruning", false, "Enable pod pruning for pod informer. "+
+		"Enabling this can reduce memory usage of the pod informer, but will disable pod propagation.")
 	o.addKlogFlags(flags)
 }
 
--- a/pkg/controllers/util/federatedclient/client.go
+++ b/pkg/controllers/util/federatedclient/client.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/sync/semaphore"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -63,6 +64,9 @@ type federatedClientFactory struct {
 	dynamicClientsetCache map[string]dynamicclient.Interface
 	kubeInformerCache     map[string]kubeinformer.SharedInformerFactory
 	dynamicInformerCache  map[string]dynamicinformer.DynamicSharedInformerFactory
+
+	availablePodListers *semaphore.Weighted
+	enablePodPruning    bool
 }
 
 func NewFederatedClientsetFactory(
@@ -71,6 +75,8 @@ func NewFederatedClientsetFactory(
 	informer fedcorev1a1informers.FederatedClusterInformer,
 	fedSystemNamespace string,
 	baseRestConfig *rest.Config,
+	maxPodListers int64,
+	enablePodPruning bool,
 ) FederatedClientFactory {
 	factory := &federatedClientFactory{
 		mu:                    sync.RWMutex{},
@@ -86,6 +92,10 @@ func NewFederatedClientsetFactory(
 		dynamicClientsetCache: map[string]dynamicclient.Interface{},
 		kubeInformerCache:     map[string]kubeinformer.SharedInformerFactory{},
 		dynamicInformerCache:  map[string]dynamicinformer.DynamicSharedInformerFactory{},
+		enablePodPruning:      enablePodPruning,
+	}
+	if maxPodListers > 0 {
+		factory.availablePodListers = semaphore.NewWeighted(maxPodListers)
 	}
 	factory.handle, _ = informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
@@ -310,6 +320,7 @@ func (f *federatedClientFactory) processQueueItem(ctx context.Context) {
 	f.dynamicClientsetCache[name] = dynamicClientset
 	f.kubeInformerCache[name] = kubeInformerFactory
 	f.dynamicInformerCache[name] = dynamicInformerFactory
+	addPodInformer(ctx, kubeInformerFactory, kubeClientset, f.availablePodListers, f.enablePodPruning)
 	f.mu.Unlock()
 }
 
--- a/pkg/controllers/util/federatedclient/podinformer.go
+++ b/pkg/controllers/util/federatedclient/podinformer.go
@@ -1,0 +1,135 @@
+/*
+Copyright 2023 The KubeAdmiral Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package federatedclient
+
+import (
+	"context"
+	"time"
+
+	"golang.org/x/sync/semaphore"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/informers"
+	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+func addPodInformer(ctx context.Context,
+	informer informers.SharedInformerFactory,
+	client kubeclient.Interface,
+	podListerSemaphore *semaphore.Weighted,
+	enablePodPruning bool) {
+	informer.InformerFor(&corev1.Pod{}, func(k kubeclient.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
+		return cache.NewSharedIndexInformer(
+			podListerWatcher(ctx, client, podListerSemaphore, enablePodPruning),
+			&corev1.Pod{},
+			resyncPeriod,
+			cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+		)
+	})
+}
+
+func podListerWatcher(ctx context.Context,
+	client kubeclient.Interface,
+	semaphore *semaphore.Weighted,
+	enablePodPruning bool) cache.ListerWatcher {
+	return &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			if semaphore != nil {
+				if err := semaphore.Acquire(ctx, 1); err != nil {
+					return nil, err
+				}
+				defer semaphore.Release(1)
+			}
+			pods, err := client.CoreV1().Pods(corev1.NamespaceAll).List(ctx, options)
+			if err != nil {
+				return nil, err
+			}
+			if enablePodPruning {
+				for i := range pods.Items {
+					prunePod(&pods.Items[i])
+				}
+			}
+			return pods, nil
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			watcher, err := client.CoreV1().Pods(corev1.NamespaceAll).Watch(ctx, options)
+			if err != nil {
+				return nil, err
+			}
+			if !enablePodPruning {
+				return watcher, nil
+			}
+
+			// It's easy for a consumer to add buffering via an extra
+			// goroutine/channel, but impossible for them to remove it,
+			// so nonbuffered is better. -- from watch.NewStreamWatcher
+			proxyCh := make(chan watch.Event)
+			proxyWatcher := watch.NewProxyWatcher(proxyCh)
+			go func() {
+				defer watcher.Stop()
+				// Closing proxyCh will notify the reflector to stop the current
+				// watching cycle and then restart the list and watch.
+				defer close(proxyCh)
+				for {
+					select {
+					case <-proxyWatcher.StopChan():
+						return
+					case event, ok := <-watcher.ResultChan():
+						if !ok {
+							// the watcher has been closed, stop the proxy
+							return
+						}
+						if pod, ok := event.Object.(*corev1.Pod); ok {
+							prunePod(pod)
+						}
+						proxyCh <- event
+					}
+				}
+			}()
+			return proxyWatcher, nil
+		},
+	}
+}
+
+func prunePod(pod *corev1.Pod) {
+	containers := make([]corev1.Container, len(pod.Spec.Containers))
+	initContainers := make([]corev1.Container, len(pod.Spec.InitContainers))
+	for i := range pod.Spec.Containers {
+		containers[i] = corev1.Container{Resources: pod.Spec.Containers[i].Resources}
+	}
+	for i := range pod.Spec.InitContainers {
+		initContainers[i] = corev1.Container{Resources: pod.Spec.InitContainers[i].Resources}
+	}
+	*pod = corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            pod.Name,
+			Namespace:       pod.Namespace,
+			Generation:      pod.Generation,
+			ResourceVersion: pod.ResourceVersion,
+			UID:             pod.UID,
+		},
+		Spec: corev1.PodSpec{
+			NodeName:       pod.Spec.NodeName,
+			Overhead:       pod.Spec.Overhead,
+			Containers:     containers,
+			InitContainers: initContainers,
+		},
+	}
+}
--- a/pkg/controllers/util/federatedclient/podinformer_test.go
+++ b/pkg/controllers/util/federatedclient/podinformer_test.go
@@ -1,0 +1,320 @@
+/*
+Copyright 2023 The KubeAdmiral Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package federatedclient
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"golang.org/x/sync/semaphore"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/rand"
+	kubeinformer "k8s.io/client-go/informers"
+	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	fakecorev1 "k8s.io/client-go/kubernetes/typed/core/v1/fake"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/util"
+)
+
+type fakePodsWithConcurrencyLimit struct {
+	corev1client.PodInterface
+	concurrentLists *atomic.Int64
+}
+
+func (c *fakePodsWithConcurrencyLimit) List(ctx context.Context, opts metav1.ListOptions) (result *corev1.PodList, err error) {
+	cur := c.concurrentLists.Add(-1)
+	defer c.concurrentLists.Add(1)
+	if cur < 0 {
+		panic("limit exceeded")
+	}
+	return c.PodInterface.List(ctx, opts)
+}
+
+type fakeCoreV1WithConcurrencyLimit struct {
+	fakecorev1.FakeCoreV1
+	concurrentLists *atomic.Int64
+}
+
+func (c *fakeCoreV1WithConcurrencyLimit) Pods(namespace string) corev1client.PodInterface {
+	return &fakePodsWithConcurrencyLimit{c.FakeCoreV1.Pods(namespace), c.concurrentLists}
+}
+
+var _ kubeclient.Interface = &fakeClientsetWithConcurrencyLimit{}
+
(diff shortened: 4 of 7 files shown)


pkg/controllers/util/federatedclient/client.go
```diff
@@ -63,6 +63,8 @@ type federatedClientFactory struct {
 	dynamicClientsetCache map[string]dynamicclient.Interface
 	kubeInformerCache     map[string]kubeinformer.SharedInformerFactory
 	dynamicInformerCache  map[string]dynamicinformer.DynamicSharedInformerFactory
+
+	availablePodListers int32
```
SOF3, line 67: `atomic.Int32`

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+		)
+	})
+}
+
+func (f *federatedClientFactory) podListerWatcher(client kubeclient.Interface) cache.ListerWatcher {
+	return &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			cur := atomic.LoadInt32(&f.availablePodListers)
+			if cur == 0 {
+				return nil, fmt.Errorf("do not have enought lister")
```
SOF3, line 34: What happens when this returns error? Does `Reflector` backoff-retry? Is it better to simply use use a `make(chan struct{}, maxPodListers)` as the semaphore?

SOF3 (changes requested):

pkg/controllers/util/federatedclient/podinformer.go
JackZxj: - What happens when this returns error? Does `Reflector` backoff-retry?
Just logs an error. It will retry, but not sure backoff or just poll. You can see that in test case `4 clusters, limit 2`.

- Is it better to simply use use a `make(chan struct{}, maxPodListers)` as the semaphore?
Do you mean block it when there is not enough lister? Like this:
```go
ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
        ch <- struct{}{}
        defer <-ch
        // do list here...
}
```

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			if cur > 0 {
+				if !atomic.CompareAndSwapInt32(&f.availablePodListers, cur, cur-1) {
+					return nil, fmt.Errorf("failed to compete for lister")
+				}
+				defer atomic.AddInt32(&f.availablePodListers, 1)
```
SOF3, line 40: So we immediately release the semaphore after this function returns, but the list is not GC'ed yet because it is yet to be consumed by the caller of this list function. Would this defeat the memory surge optimization we are trying to achieve?

cmd/controller-manager/app/options/options.go
```diff
@@ -109,6 +111,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		"The maximum amount of time to wait for a new cluster to join the federation before timing out.",
 	)
 
+	flags.Int32Var(&o.MaxPodListers, "max-pod-listers", -1, "The maximum number of pod lister used to list pods from every member clusters at the same time. "+
+		"Should be a non-zero integer. A negative number means unlimited, but may increase the instantaneous memory usage.")
```
gary-lgy, line 115:
```suggestion
	flags.Int32Var(&o.MaxPodListers, "max-pod-listers", -1, "The maximum number of concurrent pod listing requests to member clusters. "+
		"A negative number means unlimited, but may increase the instantaneous memory usage.")
```

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+func (f *federatedClientFactory) addPodInformer(informer informers.SharedInformerFactory, client kubeclient.Interface) {
+	informer.InformerFor(&corev1.Pod{}, func(k kubeclient.Interface, duration time.Duration) cache.SharedIndexInformer {
+		return cache.NewSharedIndexInformer(
+			f.podListerWatcher(client),
+			&corev1.Pod{},
+			0,
+			cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+		)
+	})
+}
```
gary-lgy, line 27:
```suggestion
func (f *federatedClientFactory) addPodInformer(informer informers.SharedInformerFactory, client kubeclient.Interface) {
	informer.InformerFor(&corev1.Pod{}, func(k kubeclient.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
		return cache.NewSharedIndexInformer(
			f.podListerWatcher(client),
			&corev1.Pod{},
			resyncPeriod,
			cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
		)
	})
}
```

pkg/controllers/util/federatedclient/podinformer.go
gary-lgy: Agree with @SOF3 here, we should manage the concurrency here explicitly instead of relying on `Reflector` backoff retry logic, which may not be suited for a use case like this. I think the channel method @JackZxj suggested would work; alternatively we can use `sync.Semaphore`.

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+			if cur == 0 {
+				return nil, fmt.Errorf("do not have enought lister")
+			}
+			if cur > 0 {
+				if !atomic.CompareAndSwapInt32(&f.availablePodListers, cur, cur-1) {
+					return nil, fmt.Errorf("failed to compete for lister")
+				}
+				defer atomic.AddInt32(&f.availablePodListers, 1)
+			}
+			pods, err := client.CoreV1().Pods(corev1.NamespaceAll).List(context.TODO(), options)
```
gary-lgy, line 42: Might want to silence the lint here: `//nolint: contextcheck`

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+			if err != nil {
+				return nil, err
+			}
+			for i, pod := range pods.Items {
+				pods.Items[i] = simplifyPod(pod)
+			}
```
gary-lgy, line 48: We could save 2 copies here.
```suggestion
			for i  := range pods.Items {
				simplifyPod(&pod.Items[i])
			}
```

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+			proxyWatcher := watch.NewProxyWatcher(proxyCh)
+			go func() {
+				for {
+					select {
+					case <-proxyWatcher.StopChan():
+						watcher.Stop()
+					case event, ok := <-watcher.ResultChan():
+						if !ok {
+							// the watcher has been closed, stop the proxy
+							close(proxyCh)
```
gary-lgy, line 70: Is it more correct to use `proxyWatcher.Close()` here?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/util"
+)
+
+func setupFakeClient(numPods int) (pods []runtime.Object, client kubernetes.Interface) {
+	pods = make([]runtime.Object, numPods)
+	for i := 1; i <= numPods; i++ {
```
gary-lgy, line 27: nit: is it necessary to use 1-based indexing here?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+			ch <- 1
+		}
+		if pod.Labels != nil {
+			return fmt.Errorf("expect no label for pod, but got labels: %v", pod.Labels)
+		}
+		return nil
+	}
+
+	for _, tt := range tests {
+		for i := 0; i < tt.clusters; i++ {
```
gary-lgy, line 124: This initialisation block should be inside `t.Run` as well, so any panics can be linked to the subtest.

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func Test_federatedClientFactory_podInformer(t *testing.T) {
+	type fields struct {
```
gary-lgy, line 74: This struct's naming is a bit confusing and I don't see why we need it. The 2 maps can be initialised in `t.Run` and `availablePodListers` can be a field under the anonymous test case struct.

cmd/controller-manager/app/options/options.go
```diff
@@ -109,6 +111,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		"The maximum amount of time to wait for a new cluster to join the federation before timing out.",
 	)
 
+	flags.Int32Var(&o.MaxPodListers, "max-pod-listers", -1, "The maximum number of pod lister used to list pods from every member clusters at the same time. "+
```
gary-lgy, line 114: There should be a validation to ensure the value is not 0.

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+							if err := handlerFn(obj, testPod.Name, deleteCh); err != nil {
+								t.Errorf("unexpected err: %v", err)
+							}
+						},
+					})
+
+					if testPod, err = f.kubeClientsetCache[clusters[i]].CoreV1().Pods(corev1.NamespaceAll).Create(
+						ctx, testPod, metav1.CreateOptions{}); err != nil {
+						t.Errorf("Unexpected error when creating pod: %v", err)
+						return
```
gary-lgy, line 198:
```suggestion
						t.Fatalf("Unexpected error when creating pod: %v", err)
```

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+							corev1.ResourceCPU: resource.MustParse("10m"),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func Test_federatedClientFactory_podInformer(t *testing.T) {
```
gary-lgy, line 73: This test verifies whether listing and watching works as expected. Do we need another test to verify the concurrency control is taking effect?

gary-lgy, line 73: It looks like the pod informer can be tested in isolation from `federatedClientFactory`. If possible, I think it's preferable to coupling the two.

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,109 @@
...
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			watcher, err := client.CoreV1().Pods(corev1.NamespaceAll).Watch(context.TODO(), options)
+			if err != nil {
+				return nil, err
+			}
+
+			// It's easy for a consumer to add buffering via an extra
+			// goroutine/channel, but impossible for them to remove it,
+			// so nonbuffered is better. -- from watch.NewStreamWatcher
+			proxyCh := make(chan watch.Event)
```
gary-lgy, line 60: It's simpler to just specify a `Transform` function (i.e.`simplifyPod`) and return `watcher` as is. Do you think this approach might work?

gary-lgy: For LIST, it might be necessary to explicitly simplify the pods before returning to (hopefully) expedite GC, but WATCH should not have the same concern. @SOF3 WDYT?

pkg/controllers/util/federatedclient/client.go
```diff
@@ -311,6 +315,7 @@ func (f *federatedClientFactory) processQueueItem(ctx context.Context) {
 	f.kubeInformerCache[name] = kubeInformerFactory
 	f.dynamicInformerCache[name] = dynamicInformerFactory
 	f.mu.Unlock()
+	f.addPodInformer(kubeInformerFactory, kubeClientset)
```
limhawjia, line 318: this function should be called within the critical section above (i.e. before `f.mu.Unlock()` on line 317).  if not there may be a chance that the shared pod informer is accessed by another goroutine before this function is called, causing the default pod informer to be constructed instead

pkg/controllers/util/federatedclient/podinformer.go
limhawjia: +1 to blocking with a semaphore

pkg/controllers/util/federatedclient/podinformer.go
limhawjia: +1

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,236 @@
...
+			podNum:            10000,
+			expectCacheSynced: true,
+			fields: fields{
+				kubeClients:         map[string]kubeclient.Interface{},
+				informers:           map[string]kubeinformer.SharedInformerFactory{},
+				availablePodListers: 2,
+			},
+		},
+		{
+			name:              "4 clusters, limit 0",
```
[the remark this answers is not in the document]
limhawjia: +1

pkg/controllers/util/federatedclient/podinformer_test.go
limhawjia: apart from concurrency control, should we also test if the field pruning/transformation is taking effect?

pkg/controllers/util/federatedclient/podinformer.go
limhawjia: i think both approaches are fine. but since we are already transforming in the list function, maybe we can forgo using the `Transform` function all together to save some cpu cycles?

pkg/controllers/util/federatedclient/podinformer_test.go
gary-lgy: The `handlerFn` tests whether labels are removed, but of course more comprehensive coverage would be good.

cmd/controller-manager/app/options/options.go
```diff
@@ -53,6 +53,9 @@ type Options struct {
 	NSAutoPropExcludeRegexp string
 	CreateCRDsForFTCs       bool
 	ClusterJoinTimeout      time.Duration
+
+	MaxPodListers        int64
+	EnablePodSimplifying bool
```
gary-lgy, line 58:
```suggestion
	EnablePodSimplification bool
```

cmd/controller-manager/app/options/options.go
```diff
@@ -109,6 +112,10 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		"The maximum amount of time to wait for a new cluster to join the federation before timing out.",
 	)
 
+	flags.Int64Var(&o.MaxPodListers, "max-pod-listers", 0, "The maximum number of concurrent pod listing requests to member clusters. "+
+		"A non-positive number means unlimited, but may increase the instantaneous memory usage.")
+	flags.BoolVar(&o.EnablePodSimplifying, "enable-pod-simplifying", false, "Enable pod simplifying for pod informer. "+
+		"Enabling this can reduce memory usage of pod informer, but cannot coexist with enabling pod scheduling.")
```
gary-lgy, line 118:
```suggestion
		"Enabling this can reduce memory usage of the pod informer, but will disable pod propagation.")
```

cmd/controller-manager/app/options/options.go
gary-lgy: Pls rmb to change all instances of this flag

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,123 @@
...
+func podListerWatcher(ctx context.Context, client kubeclient.Interface, semaphore *semaphore.Weighted, enableSimplifying bool) cache.ListerWatcher {
+	return &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			if semaphore != nil {
+				if err := semaphore.Acquire(ctx, 1); err != nil {
+					return nil, err
+				}
+				defer semaphore.Release(1)
+			}
+			pods, err := client.CoreV1().Pods(corev1.NamespaceAll).List(ctx, options)
```
gary-lgy, line 41: The client-go informers use `context.TODO()`. Intuitively I think it does make sense to shutdown the listwatch when the federatedclient is shutting down, but I can't be sure. @JackZxj @limhawjia @SOF3 @mrlihanbo WDYT?

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,123 @@
...
+					case <-proxyWatcher.StopChan():
+						return
+					case event, ok := <-watcher.ResultChan():
+						if !ok {
+							// the watcher has been closed, stop the proxy
+							return
+						}
+						switch event.Type {
+						case watch.Modified:
+							// Ignore the MODIFIED event, because the request and limit cannot
```
gary-lgy, line 82: This comment only applied to resource collection. With this implementation, the pod informer can be used for other purposes like auto-migration (#9) and resource compensation (#21) which used to rely on pod listing due to the absence of a memory-efficient pod caching mechanism. Meanwhile, handling update events should only increase the CPU but not the memory usage, and the former is currently not a concern.
Is there any other reason for ignoring update events? If not, I think it's ok to handle them.

cmd/controller-manager/app/options/options.go
SOF3: "simplification" is too long, what about "pruning"?

pkg/controllers/util/federatedclient/podinformer.go
SOF3: I think we can reorganize all context usages in a separate PR; the scope is too big and doesn't need to be implemented here.

pkg/controllers/util/federatedclient/podinformer.go
```diff
@@ -0,0 +1,144 @@
...
+						proxyCh <- event
+					}
+				}
+			}()
+			return proxyWatcher, nil
+		},
+	}
+}
+
+func simplifyPod(pod *corev1.Pod) {
```
SOF3, line 120: Should we make this function (as well as the container processing part) customizable additively from the caller so that other components can add their own requested fields too?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+var _ kubeclient.Interface = &admiralFakeClientset{}
+
+type admiralFakeClientset struct {
+	*fake.Clientset
+	concurrentLists *atomic.Int64
+}
+
+func (c *admiralFakeClientset) CoreV1() corev1client.CoreV1Interface {
```
gary-lgy, line 77:
```suggestion
func (c *fakeClientsetWithConcurrencyLimit) CoreV1() typedcorev1.CoreV1Interface {
```
`admiralFakeClientset` could mean the "fake clientset for the admiral api".

pkg/controllers/util/federatedclient/podinformer.go
gary-lgy: Agree that context usage should be fixed in a separate PR. I'm just wondering whether using `ctx` instead of `context.TODO` here is correct. Using `ctx` would mean the `ListWatch` would shutdown when `ctx` is cancelled.

cmd/controller-manager/app/options/options.go
gary-lgy: Prefer "pruning"

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+	type memberCluster struct {
+		name              string
+		client            kubeclient.Interface
+		informer          kubeinformer.SharedInformerFactory
+		enableSimplifying bool
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			defer func() {
```
gary-lgy, line 195: What's the benefit of recovering instead of letting it panic?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+							corev1.ResourceCPU: resource.MustParse("10m"),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+//nolint:gocyclo
```
gary-lgy, line 142: What was the complexity gocyclo reported? Any chance we can address the lint instead of suppressing it?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+					defer cancel()
+
+					// test lister
+					if !cache.WaitForNamedCacheSync(clusters[i].name, ctx.Done(), podsSynced) {
+						if tt.expectCacheSynced {
+							t.Errorf("%s should be synced, but it was not\n", clusters[i].name)
+						}
+						return
+					}
+					if !tt.expectCacheSynced {
```
gary-lgy, line 248: Would this ever happen? Why do we need `expectCacheSynced` in the test case?

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+						t.Errorf("%s should not be synced, but it was\n", clusters[i].name)
+						return
+					}
+					pods, err := podLister.List(labels.Everything())
+					if err != nil {
+						t.Errorf("%s: unexpected error when listing pod: %v\n", clusters[i].name, err)
```
gary-lgy, line 254:
```suggestion
						t.Fatalf("%s: unexpected error when listing pod: %v\n", clusters[i].name, err)
```
Generally when an error occurs, there's little reason to continue with the rest of the test case.

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+							return fmt.Errorf("expected to handle pod, but got: %v", obj)
+						}
+						if pod.Name == watchPodName {
+							ch <- 1
+						}
+						if clusters[i].enableSimplifying && pod.Labels != nil {
+							return fmt.Errorf("expect no label for pod, but got labels: %v", pod.Labels)
+						}
+						if !clusters[i].enableSimplifying && pod.Labels == nil {
+							return fmt.Errorf("expected to get lables form pod, but got nil")
```
gary-lgy, line 276:
```suggestion
							return fmt.Errorf("expected to get lables from pod, but got nil")
```

pkg/controllers/util/federatedclient/podinformer_test.go
gary-lgy: Preferably all `Errorf` should be replaced with `Fatalf`

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+						}
+						return nil
+					}
+					addCh, updateCh, deleteCh := make(chan int, 1), make(chan int, 1), make(chan int, 1)
```
gary-lgy, line 280:
```suggestion
					addCh, updateCh, deleteCh := make(chan struct{}), make(chan struct{}), make(chan struct{})
```

pkg/controllers/util/federatedclient/podinformer_test.go
```diff
@@ -0,0 +1,346 @@
...
+					watchPodName := fmt.Sprintf("watch-pod-%d", rand.Intn(tt.podNum))
+					watchPodNamespace := "watch"
+					watchPod := newPod(watchPodNamespace, watchPodName)
+					handlerFn := func(obj interface{}, ch chan int) error {
+						pod, ok := obj.(*corev1.Pod)
+						if !ok {
+							return fmt.Errorf("expected to handle pod, but got: %v", obj)
+						}
+						if pod.Name == watchPodName {
+							ch <- 1
```
gary-lgy, line 270:
```suggestion
							close(ch)
```

pkg/controllers/util/federatedclient/podinformer.go
gary-lgy: This is probably out of scope. We can add this later as required.

gary-lgy (approved):

SOF3 (approved):

Issue #10, opened 2023-04-03
Refactor controllers to follow logging conventions


limhawjia:
The following controllers have to be refactored to follow the new logging conventions in the code style [guide](https://github.com/kubewharf/kubeadmiral/blob/main/docs/code-style.md#logging-conventions). 

This will reduce the amount of noise produced in the controller-manager's logs and make the logs much more readable.

- [x] FederatedCluster controller
- [x] Scheduler
- [x] Sync controller
- [x] Federate controller
- [x] Status controller
- [x] StatusAggregator controller 
- [x] Follower controller 
- [x] PolicyRC controller

limhawjia:
This is finally finished, thanks @qclc and @wy-lucky

PR #48, opened 2023-04-28
Webhook plugin integration

gary-lgy:
Notable changes:

Features:
- Allow webhook plugins to be referenced and used by `SchedulerProfile` (with integration tests)
- webhook plugins now handle non-200 responses correctly (with unit tests)

Fixes:
- extract workload template correctly when creating scheduling unit (with unit tests)

Refactors:
- split scheduler's `reconcile` function into smaller functions
- standardise logging in `generic_scheduler.go`
- scheduling `Algorithm` no longer holds a reference to cluster store; clusters are passed explicitly into `Schedule`

Docs:
- Add sample scheduler webhook

### DIFF
--- a/pkg/controllers/scheduler/scheduler.go
+++ b/pkg/controllers/scheduler/scheduler.go
@@ -332,7 +332,7 @@ func (s *Scheduler) prepareToSchedule(
 					"object propagation policy %s not found",
 					policyKey.String(),
 				)
-				return nil, nil, nil, &worker.Result{Success: false, RequeueAfter: nil}
+				return nil, nil, nil, &worker.StatusAllOK
 			}
 			return nil, nil, nil, &worker.StatusError
 		}
--- a/pkg/controllers/scheduler/scheduler.go
+++ b/pkg/controllers/scheduler/scheduler.go
@@ -96,6 +96,7 @@ type Scheduler struct {
 }
 
 func NewScheduler(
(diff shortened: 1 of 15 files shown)


pkg/controllers/scheduler/scheduler.go
```diff
@@ -268,23 +332,44 @@ func (s *Scheduler) reconcile(qualifiedName common.QualifiedName) (status worker
...
+				s.eventRecorder.Eventf(
+					fedObject,
+					corev1.EventTypeWarning,
+					EventReasonScheduleFederatedObject,
+					"failed to schedule object: %v",
+					fmt.Errorf("failed to get scheduling profile %s: %w", profileName, err),
+				)
+
+				if apierrors.IsNotFound(err) {
+					return nil, nil, nil, &worker.StatusAllOK
```
limhawjia, line 355: should we return `&worker.Result{Success: false, RequeueAfter: nil}` instead of `&worker.StatusAllOK` to be consistent with line 335 above?

gary-lgy: Changed line 335 to return `&worker.StatusAllOK` as well. We should remove the `Success` field in `worker.Result` eventually.

limhawjia (approved):
LGTM

limhawjia:
completes scheduler task in #10

PR #49, opened 2023-04-28
refactor: logging in federate controller

limhawjia:
Refactors the federate controller to follow the new logging conventions

### DIFF
--- a/pkg/controllers/federate/controller.go
+++ b/pkg/controllers/federate/controller.go
@@ -52,6 +52,8 @@ import (
 )
 
 const (
+	FederateControllerName = "federate-controller"
+
 	EventReasonCreateFederatedObject = "CreateFederatedObject"
 	EventReasonUpdateFederatedObject = "UpdateFederatedObject"
 
@@ -101,13 +103,14 @@ func NewFederateController(
 	fedSystemNamespace string,
 ) (*FederateController, error) {
 	controllerName := fmt.Sprintf("%s-federate-controller", typeConfig.GetFederatedType().Name)
+	logger := klog.LoggerWithValues(klog.Background(), "controller", FederateControllerName, "ftc", typeConfig.Name)
 
 	c := &FederateController{
 		typeConfig:         typeConfig,
 		name:               controllerName,
 		fedSystemNamespace: fedSystemNamespace,
 		metrics:            metrics,
-		logger:             klog.LoggerWithName(klog.Background(), controllerName),
+		logger:             logger,
 	}
 
 	c.worker = worker.NewReconcileWorker(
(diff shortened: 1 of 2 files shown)


pkg/controllers/federate/controller.go
```diff
@@ -97,13 +99,14 @@ func NewFederateController(
 	fedSystemNamespace string,
 ) (*FederateController, error) {
 	controllerName := fmt.Sprintf("%s-federate-controller", typeConfig.GetFederatedType().Name)
+	logger := klog.LoggerWithValues(klog.Background(), "controller", FederateControllerName, "ftc", typeConfig.Name)
```
gary-lgy, line 102: Should we pass a parent logger into the constructor to allow top-level configuraitons?

pkg/controllers/federate/controller.go
```diff
@@ -164,43 +167,42 @@ func (c *FederateController) Run(ctx context.Context) {
 
 func (c *FederateController) reconcile(qualifiedName common.QualifiedName) (status worker.Result) {
 	_ = c.metrics.Rate("federate.throughput", 1)
-	key := qualifiedName.String()
-	keyedLogger := c.logger.WithValues("control-loop", "reconcile", "key", key)
+	logger := c.logger.WithValues("object", qualifiedName.String())
+	ctx := klog.NewContext(context.Background(), logger)
```
gary-lgy, line 171:
```suggestion
	ctx := klog.NewContext(context.TODO(), logger)
```

pkg/controllers/federate/controller.go
```diff
@@ -387,25 +389,25 @@ func (c *FederateController) handleTerminatingSourceObject(ctx context.Context,
...
 
-	logger.V(6).Info("Generating federated object from source object")
+	logger.V(2).Info("Generating federated object from source object")
 	fedObject, err := newFederatedObjectForSourceObject(c.typeConfig, sourceObject)
 	if err != nil {
 		return fmt.Errorf("failed to generate federated object from source object: %w", err)
 	}
 
-	logger.V(6).Info("Set pending controllers on federated object")
+	logger.V(2).Info("Set pending controllers on federated object")
```
gary-lgy, line 405: This log feels a bit unnecessary. Can we remove it

pkg/controllers/federate/controller.go
```diff
@@ -423,17 +425,17 @@ func (c *FederateController) handleExistingFederatedObject(
 ) (bool, error) {
 	logger := klog.FromContext(ctx)
 
-	logger.V(6).Info("Checking if federated object needs update")
+	logger.V(2).Info("Checking if federated object needs update")
```
gary-lgy, line 428: At V(2), this line will be printed too many times. Can we bump it to V(3) or remove it?

pkg/controllers/federate/controller.go
limhawjia: what top level configurations do we need? i think this is ok for now, in the future we may retrieve a parent logger from the context passed to the reconcile function (once the worker is refactored)

gary-lgy: Configurations like log sink, level, etc.

gary-lgy (approved):

PR #57, opened 2023-05-04
docs: make quickstart clearer

gary-lgy:
* comment out dead links
* fix wrong headers
* make the flow from `Installation` to `Managing Kubernetes resources` smoother
* remove the need to create temporary files

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -2,18 +2,14 @@
 
 This guide will cover:
 
-* [Installation](#installation): different ways to setup and run the KubeAdmiral control plane.
+* [Installation](#installation): running KubeAdmiral against a local federation of fake clusters.
 * [Managing Kubernetes resources](#managing-kubernetes-resources): basic example of propgating a Deployment resource.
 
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -267,9 +268,3 @@ $ kubectl get fdeploystatus echo-server -oyaml | yq .clusterStatus
 ```
```
qclc, line 268: The `yq` command tool that `kubectl get fdeploystatus echo-server -oyaml | yq .clusterStatus` depends on in step 5 needs to be installed in advance. How about adding a prompt in this step?

gary-lgy: Not sure if we should require users to install `yq` just to filter out some lines... I think we can remove the ` | yq .clusterStatus` part and let users see the full output. @limhawjia WDYT?

gary-lgy: Fixed, @limhawjia @SOF3 @qclc PTAL

limhawjia (approved):

PR #61, opened 2023-05-09
feat: support pod propagation

limhawjia:
This PR supports the federation of pods using KubeAdmiral by introducing the following:

1. Add a new FTC for pods
2. Implement drop and retain logic for pod fields before syncing

Additionally:

1. The kube-scheduler is now disabled by the `dev-up.sh` script for the host cluster.
2. `dev-up.sh` now supports running 0 member clusters


config/sample/host/01-ftc.yaml
```diff
@@ -692,3 +692,37 @@ spec:
     fields:
       - metadata.creationTimestamp
       - status
+---
+apiVersion: core.kubeadmiral.io/v1alpha1
```
gary-lgy, line 696: Should we include it by default? This means all users will start caching pods.

config/sample/host/01-ftc.yaml
```diff
@@ -692,3 +692,37 @@ spec:
...
+    kind: FederatedPodStatus
+    pluralName: federatedpodstatuses
+    scope: Namespaced
+    version: v1alpha1
+  controllers:
+    - - kubeadmiral.io/global-scheduler
+    - - kubeadmiral.io/overridepolicy-controller
+  statusCollection:
+    fields:
+      - status
```
gary-lgy, line 728: Should we also collect spec.nodeName?

config/sample/host/01-ftc.yaml
SOF3: Also would it cause problems with static pods?

config/sample/host/01-ftc.yaml
SOF3: What happens when pod is evicted in the member?

config/sample/host/01-ftc.yaml
limhawjia: @gary-lgy probably not 🙂, i will move it to a separate file

limhawjia:
> @limhawjia pls make sure to reconcile with #46, especially the interaction between pod pruning and pod propagation

The sync controller (along with other subcontrollers that will require the whole pod object) currently do not use shared informers, so it will not be affected by #46. However, I opened a new issue #64 as a reminder in the event that we do refactor these subcontrollers to use shared informers.

limhawjia:
Pod propagation is only supported if the `BoundServiceAccountTokenVolume` is enabled (enabled by default from 1.21 onwards).

Without `BoundServiceAccountTokenVolume`, erviceaccount volumes can only be identified by checking if the volume has a secret volumeSource that references the secret containing the serviceaccount token. This would require us to send multiple requests to the apiserver or add new informers to the sync controller, adding a lot of overhead. As a result, we will not be supporting this for now.

pkg/controllers/sync/dispatch/retain.go
```diff
@@ -251,6 +263,256 @@ func retainPersistentVolumeClaimFields(
...
+func copyUnstructuredField(srcObj, destObj *unstructured.Unstructured, fields ...string) error {
+	value, exists, err := unstructured.NestedFieldNoCopy(srcObj.Object, fields...)
+	if err != nil || !exists {
+		return nil
+	}
+
+	return unstructured.SetNestedField(destObj.Object, value, fields...)
+}
+
+// containsServiceAccountVolume checks if the given pod contains a serviceaccount volume as defined by the
```
gary-lgy, line 355: Function name the comment is outdated.

pkg/controllers/sync/dispatch/retain.go
```diff
@@ -251,6 +263,256 @@ func retainPersistentVolumeClaimFields(
...
+	if err := retainContainers(desiredInitContainers, clusterInitContainers); err != nil {
+		return err
+	}
+	if err := unstructured.SetNestedSlice(desiredObj.Object, desiredInitContainers, "spec", "initContainers"); err != nil {
+		return err
+	}
+
+	// The tolerations field is also defaulted by an admission plugin. However, we do not need to explicitly retain
+	// the default tolerations due to 2 reasons:
+	// 1. The tolerations field is mutable so any updates will not reuslt in an error
```
gary-lgy, line 331:
```suggestion
	// 1. The tolerations field is mutable so any updates will not result in an error
```

config/sample/extra/pod-ftc.yaml
gary-lgy, line 1: Should this file be under `config/sample/host/extra`?

pkg/controllers/sync/dispatch/retain.go
```diff
@@ -251,6 +263,231 @@ func retainPersistentVolumeClaimFields(
...
+			copy(desiredVolumes[idx+1:], desiredVolumes[idx:])
+			desiredVolumes[idx] = volume
+
+			if err := unstructured.SetNestedSlice(desiredObj.Object, desiredVolumes, "spec", "volumes"); err != nil {
+				return err
+			}
+		}
+	}
+
+	desiredContainers, _, _ := unstructured.NestedSlice(desiredObj.Object, "spec", "containers") // this is a deepcopy
```
gary-lgy, line 324: Why do we not need to handle err here?

limhawjia: 1. when err is returned, desiredContainers will be `nil` which is handled by `retainContainerFields`. 
2. `unstructured.NestedSlice` returns err when the value for the given field is not a `[]interface{}`, implying that the container field in the pod is corrupted for some reason - in that case, we simply ignore the field (which brings us back to 1)

gary-lgy (approved):

PR #62, opened 2023-05-11
fix: removal of unjoined cluster should not hang

gary-lgy:
Main contribution:
* Cluster that failed to join due to reachability issues will not be blocked for clean-up upon deletion
This is achieved by recording whether any effectual action was performed during joining in a new status field `JoinPerformed`. Upon deletion, we only require clean-up if `JoinPerformed` is true.
Note: a cluster that failed to join but underwent partial joining process (e.g. namespace created but rbac not created) will still require clean-up.

Other contributions:
* Refactor `FederatedCluster` API types:
  * secretRef is no longer required
  * lastTransitionTimestamp, reason, message in conditions are now value types and required
  * disallow conditions with duplicate types
  * make conditions optional
  * reword message and reason description
* Standardise contextual logging in cluster join code
* Replace the default `GinkgoLogr` to tidy up the e2e logs

### DIFF
--- a/pkg/controllers/federatedcluster/clusterjoin.go
+++ b/pkg/controllers/federatedcluster/clusterjoin.go
@@ -80,14 +80,17 @@ const (
 )
 
 func handleNotJoinedCluster(
+	ctx context.Context,
 	cluster *fedcorev1a1.FederatedCluster,
 	client fedclient.Interface,
 	kubeClient kubeclient.Interface,
 	eventRecorder record.EventRecorder,
 	fedSystemNamespace string,
-	logger klog.Logger,
 	clusterJoinTimeout time.Duration,
 ) (*fedcorev1a1.FederatedCluster, error) {
+	logger := klog.FromContext(ctx).WithValues("process", "cluster-join")
+	ctx = klog.NewContext(ctx, logger)
+
 	joinedCondition := getClusterCondition(&cluster.Status, fedcorev1a1.ClusterJoined)
 
 	// 1. check for join timeout
@@ -268,7 +271,7 @@ func handleNotJoinedCluster(
 
 	if cluster.Spec.UseServiceAccountToken {
 		logger.Info("Get and save cluster token")
-		err = getAndSaveClusterToken(cluster, kubeClient, clusterKubeClient, fedSystemNamespace, memberFedNamespace, logger)
+		err = getAndSaveClusterToken(ctx, cluster, kubeClient, clusterKubeClient, fedSystemNamespace, memberFedNamespace, logger)
 
 		if err != nil {
 			eventRecorder.Eventf(
@@ -383,6 +386,7 @@ func getOrCreateFedSystemNamespace(
 }
 
 func getAndSaveClusterToken(
+	ctx context.Context,
 	cluster *fedcorev1a1.FederatedCluster,
 	kubeClient kubeclient.Interface,
 	clusterKubeClient kubeclient.Interface,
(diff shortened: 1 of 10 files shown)


pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -708,3 +553,48 @@ func getServiceAccountToken(
...
+	cluster *fedcorev1a1.FederatedCluster,
+	client fedclient.Interface,
+	eventRecorder record.EventRecorder,
+
+	joinError error,
+
+	eventType, eventReason, eventMessage string,
+
+	conditionStatus corev1.ConditionStatus,
+	conditionReason, conditionMessage string,
```
SOF3, line 568: can we refactor this into a struct instead? this function has 11 parameters...

pkg/controllers/federatedcluster/controller.go
```diff
@@ -333,44 +337,19 @@ func handleTerminatingCluster(
...
+		clusterSecret, clusterKubeClient, err := getClusterClient(ctx, kubeClient, fedSystemNamespace, cluster)
 		if err != nil {
 			eventRecorder.Eventf(
 				cluster,
 				corev1.EventTypeWarning,
 				EventReasonHandleTerminatingClusterFailed,
-				"cluster %q secret %q not found",
-				cluster.Name,
-				clusterSecretName,
+				"Failed t get cluster client: %v",
```
SOF3, line 348:
```suggestion
				"Failed tp get cluster client: %v",
```

pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -80,232 +75,162 @@ const (
...
+	ctx context.Context,
 	cluster *fedcorev1a1.FederatedCluster,
 	client fedclient.Interface,
 	kubeClient kubeclient.Interface,
 	eventRecorder record.EventRecorder,
 	fedSystemNamespace string,
-	logger klog.Logger,
 	clusterJoinTimeout time.Duration,
-) (*fedcorev1a1.FederatedCluster, error) {
+) (*fedcorev1a1.FederatedCluster, *fedcorev1a1.ClusterCondition, error) {
```
limhawjia, line 85: why do we return the cluster condition separately from the cluster object?

pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -80,232 +75,162 @@ const (
...
-			EventReasonJoinClusterError,
-			"cluster %q secret %q not found",
-			cluster.Name,
-			clusterSecretName,
+			corev1.EventTypeWarning, EventReasonJoinClusterError, msg,
 		)
-		return cluster, fmt.Errorf("cluster secret not found: %w", err)
+		return cluster, &fedcorev1a1.ClusterCondition{
+			Status:  corev1.ConditionFalse,
+			Reason:  TokenNotObtainedReason,
```
limhawjia, line 123: should the reason here be `TokenNotObtained`? `TokenNotObtained` refers to not being able to find the service account token in the member cluster right? would it be misleading to reuse the same reason?

pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -80,232 +75,162 @@ const (
...
+		if !apierrors.IsNotFound(err) {
+			msg := fmt.Sprintf("Failed to get namespace: %v", err.Error())
+			logger.Error(err, "Failed to get namespace")
+			eventRecorder.Eventf(
+				cluster,
+				corev1.EventTypeWarning, EventReasonJoinClusterError, msg,
+			)
+			return cluster, &fedcorev1a1.ClusterCondition{
+				Status:  corev1.ConditionFalse,
+				Reason:  TokenNotObtainedReason,
```
limhawjia, line 142: Wrong reason used here? Should it be `GetOrCreateNamespaceFailed` instead?

pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -80,232 +75,162 @@ const (
...
-
-	logger.Info(fmt.Sprintf("Create or get system namespace %s in cluster", fedSystemNamespace))
-	memberFedNamespace, unjoinable, err := getOrCreateFedSystemNamespace(clusterKubeClient, fedSystemNamespace, string(cluster.UID))
-	if err != nil {
+	// Either the namespace doesn't exist or it is created by us.
+	// Clean-up on removal is required.
+	cluster.Status.JoinPerformed = true
+	// Update the condition to persist the flag before proceeding to ensure
+	// we don't lose the flag due to an update failure.
+	if cluster, err = client.CoreV1alpha1().FederatedClusters().UpdateStatus(
```
limhawjia, line 172: Should we check if the original status already has `JoinPerformed` set to true to avoid unnecessary updates?

pkg/controllers/federatedcluster/clusterjoin.go
limhawjia: Oh I noticed that `GetOrCreateNamespaceFailed` was removed... Is there a reason for this?

gary-lgy: It was a mistake to expose `GetOrCreateNamespace failed` in the first place. The only promise to the users is that we will get the service account toke *somehow*, without any guarantee on *how* exactly it is done. Any errors encountered while obtaining the token will be recorded with a `TokenNotObtained` reason.

pkg/controllers/federatedcluster/clusterjoin.go
gary-lgy: It's to reuse the logic of updating the condition conditionally. If we do it before returning, we'll have to use the function with 11 parameters in a previous commit (see @SOF3 's comment).

limhawjia: got it. i just find it rather confusing that while this function mutates certain fields, it also returns a condition that that the caller has to manually set afterwards. could we add a comment to describe this behavior?

pkg/controllers/federatedcluster/clusterjoin.go
```diff
@@ -79,233 +74,168 @@ const (
 	EventReasonClusterUnjoinable          = "ClusterUnjoinable"
 )
 
+// Processes a cluster that has not joined.
+// If a condition is returned, the cluster is updated to reflect the condition.
+// The condition must have status, reason and message set. The other fields will be defaulted.
```
limhawjia, line 79:
```suggestion
// If a condition is returned, it implies that the cluster requires a status update. The condition should also be  merged into the returned cluster object by the caller before the status update.
// The condition will have status, reason and message set. The other fields should be added by the caller.
```

limhawjia (approved):

PR #63, opened 2023-05-12
chore: gofumpt

SOF3:


pkg/controllers/follower/util_test.go
```diff
@@ -34,117 +34,225 @@ func TestGetFollowersFromPod(t *testing.T) {
...
-							Name: "Spec.InitContainers[*].EnvFrom[*].SecretRef"}}}},
+							Name: "Spec.InitContainers[*].EnvFrom[*].SecretRef",
+						},
+					},
+				}},
 				Env: []corev1.EnvVar{{
 					ValueFrom: &corev1.EnvVarSource{
 						ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
 							LocalObjectReference: corev1.LocalObjectReference{
-								Name: "Spec.InitContainers[*].Env[*].ValueFrom.ConfigMapKeyRef"}}}}, {
```
SOF3, line 80: Were these intentionally formatted like this? Is there a good reason to introduce a difference "convention"? If we are really concerned about compact code, using some sort of inline YAML would be a better idea.

gary-lgy (approved):

PR #68, opened 2023-05-17
feat: add commands for building docker images and new kubeadmiral local startup

qclc:
#### What this PR does?
1. Modify the script to support compiling binary files for multiple platforms at one time(i.e., linux/darwin, amd64/arm64);
2. Add scripts to quickly generate images for different platfroms(i.e., linux/amd64, linux/arm64);
3. Add scripts to help users quickly start a local kubeadmiral;
4. Add scripts to help users quickly clean up local kind clusters related to kubeadmiral;
5. Restructure the `kubeadmiral/hack` directory structure to make the script easier to read.

#### Does this PR introduce a user-facing change?
- Users can use the following make command to compile binary files for multiple platforms:
```bash
# Production build
#
# Args:
#   BUILD_PLATFORMS: platforms to build. e.g.: linux/amd64,darwin/amd64. The default value is the host platform.
#
# Example:
#   # compile kubeadmiral-controller-manager with the host platform
#   make build
#   # compile kubeadmiral-controller-manager with specified platforms
#   make build BUILD_PLATFORMS=linux/amd64,darwin/amd64
$ make build
```
- Users can use the following make command to generate docker images for multiple architectures:
```bash
# Build binaries and docker images.
# The supported OS is linux, and user can specify the arch type (only amd64,arm64,arm are supported)
#
# ARGS:
#   REGISTRY:    image registry, the default value is "ghcr.io/kubewharf"
#   TAG:         image tag, the default value is "latest"
#   ARCHS:       list of target architectures, e.g.:amd64,arm64,arm. The default value is host arch
#
# Examples:
#   # build images with the host arch, it will generate "ghcr.io/kubewharf/kubeadmiral-controller-manager:latest"
#   make images
#
#   # build images with amd64,arm64 arch
#   # note: If you specify multiple architectures, the image tag will be added with the architecture name,e.g.:
#   # 		"ghcr.io/kubewharf/kubeadmiral-controller-manager:latest-amd64"
#   # 		"ghcr.io/kubewharf/kubeadmiral-controller-manager:latest-arm64"
#   make images ARCHS=amd64,arm64
$ make images
```
- Users can use the following make command to quickly start and experience kubeadmiral locally
```bash
# Local up the KubeAdmiral.
#
# Args:
#   NUM_MEMBER_CLUSTERS: The number of member clusters you want to startup, default is 3, at least 1.
#   REGION:              region to build. e.g.: 'cn' means china mainland(the default value)
#
# Example:
#   # It will locally start a meta-cluster to deploy the kubeadmiral control-plane, and three member cluster clusters
#   make local-up
$ make local-up
```
- Users can clean up the local kind clusters started by the `make local-up` command with the following command
```bash
# Clean up all the local kind clusters and related kubeconfigs
$ make clean-local-cluster
```


SOF3:
Why is it necessary to run our own control plane from an existing k8s cluster, instead of just using the kind cluster setup?

qclc:
> Why is it necessary to run our own control plane from an existing k8s cluster, instead of just using the kind cluster setup?

The original local startup method(`make dev-up`) directly starts kubeadmiral-controller-manager as a process.
I understand that the above method is mainly for developers, and it is easier to debug and see log information. But the containerized deployment of kubeadmiral-controller-manager is still missing.

This PR mainly adds a new kubeadmiral startup method(`make local up`). By starting the kubeadmiral-controller-manager component as a pod, it provides users with a cloud-native way to deploy the core components of kubeadmiral.

But the kubeadmiral control plane requires kube-apiserver and kube-controller-manager with limited capabilities, making it impossible to directly deploy the kubeadmiral-controller-manager pod on the kubeadmiral control plane. Therefore, this PR chooses to deploy control plane components through another meta-cluster to realize cloud-native deployment.

SOF3:
@qclc Are you implying that this setup is intended for production setup, or just a preview setup for non-developer users? For the latter, could we just use kind/kwok for running the K8S control plane, and use docker for running the kubeadmiral control plane?

qclc:
> @qclc Are you implying that this setup is intended for production setup, or just a preview setup for non-developer users? For the latter, could we just use kind/kwok for running the K8S control plane, and use docker for running the kubeadmiral control plane?

There does not seem to be much difference between running a process in a docker container and using a process directly. The kubeadmiral-controller-manager and the k8s control plane still operate in a relatively fragmented manner. So I think just use `make dev-up` directly is fine.
This PR provides a way to directly deploy the entire control plane (including k8s control plane and admiral components) in a cloud-native manner. On this basis, kubeadmiral can be deployed to the user's existing k8s cluster with one click in the future, which can meet the potential needs of some users.
As for `make local-up`, which is equivalent to starting a local kubeadmiral in a cloud-native way, I think it provides another new local installation method.

hack/lib/deploy-kubeadmiral.sh
```diff
@@ -0,0 +1,202 @@
...
+
+  local kubeadmiral_ca=$(base64 -i "${root_ca_file_path}" | tr -d '\r\n')
+  local kubeadmiral_ca_key=$(base64 -i "${root_ca_key_path}" | tr -d '\r\n')
+
+  local TEMP_PATH=$(mktemp -d)
+
+  cp -rf "${cert_yaml_path}"/kubeadmiral-cert-secret.yaml "${TEMP_PATH}"/kubeadmiral-cert-secret-tmp.yaml
+  cp -rf "${cert_yaml_path}"/kubeconfig-secret.yaml "${TEMP_PATH}"/kubeconfig-secret-tmp.yaml
+
+  sed -i'' -e "s/{{ca_crt}}/${kubeadmiral_ca}/g" "${TEMP_PATH}"/kubeadmiral-cert-secret-tmp.yaml
```
SOF3, line 143: can we clean up this function and reuse the same replace commands for all files? furthermore, the temp path seems unnecessary since we can just pipe the `sed` output to `kubctl apply -f -` directly?

hack/lib/util.sh
```diff
@@ -0,0 +1,302 @@
...
+    GIT_TREESTATE="clean"
+  else
+    GIT_TREESTATE="dirty"
+  fi
+  BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+  LDFLAGS="-X ${KUBEADMIRAL_INFO_PKG}/version.gitVersion=${GIT_VERSION} \
+            -X ${KUBEADMIRAL_INFO_PKG}/version.gitCommit=${GIT_COMMIT_HASH} \
+            -X ${KUBEADMIRAL_INFO_PKG}/version.gitTreeState=${GIT_TREESTATE} \
+            -X ${KUBEADMIRAL_INFO_PKG}/version.buildDate=${BUILD_DATE}"
```
SOF3, line 119: where are these variables used? This package doesn't seem to exist

hack/make-rules/clean-local-cluster.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ ! $(kind get clusters | grep kubeadmiral ) ]]; then
+  echo "There is no local kind cluster related with kubeadmiral."
+else
+  kind delete clusters $(kind get clusters | grep kubeadmiral | tr -s "\r\n" " ")
+  echo "The following files will be deleted:
+  ${HOME}/.kube/kubeadmiral.config
+  ${HOME}/.kube/members.config
+  ${HOME}/.kube/kubeadmiral"
+  rm -rf ${HOME}/.kube/kubeadmiral.config ${HOME}/.kube/members.config ${HOME}/.kube/kubeadmiral
```
SOF3, line 29: why do we need the `-f` here?

Makefile
```diff
@@ -1,27 +1,105 @@
...
+# Example:
+#   # compile kubeadmiral-controller-manager with the host platform
+#   make build
+#   # compile kubeadmiral-controller-manager with specified platforms
+#   make build BUILD_PLATFORMS=linux/amd64,darwin/amd64
 .PHONY: build
 build:
-	mkdir -p output
-	go build $(BUILD_FLAGS) -o output/$(BINARY) cmd/controller-manager/main.go
+	BUILD_FLAGS=$(BUILD_FLAGS) TARGET_NAME=$(TARGET_NAME) bash hack/make-rules/build.sh $@
```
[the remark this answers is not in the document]
qclc: The `$@` is removed, please take a review.

Makefile
```diff
@@ -1,27 +1,105 @@
...
+# go build arguments
 BUILD_FLAGS :=
-BINARY := manager
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
+GOPROXY ?= $(shell go env GOPROXY)
+TARGET_NAME ?= "kubeadmiral-controller-manager"
+
+# image information
+REGISTRY?="ghcr.io/kubewharf"
```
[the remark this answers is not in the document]
qclc: The quotes are removed, please take a review.

config/deploy/controlplane/kubeadmiral-controller-manager.yaml
```diff
@@ -0,0 +1,36 @@
...
+            - /bin/kubeadmiral-controller-manager
+            - --create-crds-for-ftcs=true
+            - --kubeconfig=/etc/kubeconfig
+            - --klog-v=4
+          image: ghcr.io/kubewharf/kubeadmiral-controller-manager:latest
+          imagePullPolicy: IfNotPresent
+          volumeMounts:
+            - name: kubeconfig
+              subPath: kubeconfig
+              mountPath: /etc/kubeconfig
```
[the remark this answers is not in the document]
qclc: kubeadmiral-controller-manager does not enable profiling by default, so there is no exposed port here. Users can modify the yaml file by themselves if necessary.

hack/lib/deploy-kubeadmiral.sh
qclc: > can we clean up this function and reuse the same replace commands for all files? furthermore, the temp path seems unnecessary since we can just pipe the `sed` output to `kubctl apply -f -` directly?

This function has been cleaned up, please take a review.

hack/lib/util.sh
qclc: These parameters will be used by the go build command to pass the current code version information. The specific call location is in line 24 of `hack/lib/build.sh`.

hack/make-rules/clean-local-cluster.sh
qclc: Sometimes the file to be deleted may not exist, `-f` can skip the non-existing file without displaying any error message.

hack/lib/util.sh
SOF3: I mean, is this going to be used anywhere in the code?

SOF3 (approved):

PR #77, opened 2023-05-23
feat(retain): enhance label propagation

gary-lgy:
- don't sync labels like `kubeadmiral.io/propagation-policy-name` to members
- merge desired and observed labels during update to allow controllers in member clusters to manipulate labels
- propagate deletion of labels/annotation from template to members

### DIFF
--- a/pkg/controllers/federate/util.go
+++ b/pkg/controllers/federate/util.go
@@ -30,6 +30,7 @@ import (
 	fedcorev1a1 "github.com/kubewharf/kubeadmiral/pkg/apis/core/v1alpha1"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/common"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/nsautoprop"
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/override"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler"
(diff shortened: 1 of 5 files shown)


limhawjia (approved):

pkg/controllers/federate/util.go
```diff
@@ -215,38 +198,68 @@ var (
...
-	template map[string]string,
+	federatedAnnotations map[string]string,
+	templateAnnotations map[string]string,
 ) {
-	result := map[annotationClass]map[string]string{}
-
-	for key, value := range annotations {
-		class := classifyAnnotation(key)
+	federatedAnnotations, templateAnnotations = classifyStringMap(annotations, classifyAnnotation)
+	federatedAnnotations[common.FederatedObjectAnnotation] = "1"
```
mrlihanbo, line 236: what's the meaning of "1"?

gary-lgy: It means "true". This annotation is used to distinguish whether a federated object is created from a source object, or created manually. I agree it's not the best mechanism, but we probably shouldn't change it in this PR.

mrlihanbo: > It means "true". This annotation is used to distinguish whether a federated object is created from a source object, or created manually. I agree it's not the best mechanism, but we probably shouldn't change it in this PR.

Got it

mrlihanbo:
Thanks
/lgtm

PR #78, opened 2023-05-23
fix: cluster cache sync blocks reconcile

limhawjia:
With this PR, the following controllers will no longer block when a single cluster's cache is not synced:

* Sync
* Status
* StatusAggregator
* AutoMigration

This is done by sending a direct request to the member API server before the caches are synced.

### DIFF
--- a/pkg/controllers/federatedtypeconfig/federatedtypeconfig_controller.go
+++ b/pkg/controllers/federatedtypeconfig/federatedtypeconfig_controller.go
@@ -173,6 +173,26 @@ func (c *Controller) Run(stopChan <-chan struct{}) {
 		return
 	}
 
+	var nsFTC *fedcorev1a1.FederatedTypeConfig
+	ftcs := c.ftcStore.List()
+	for _, obj := range ftcs {
+		ftc := obj.(*fedcorev1a1.FederatedTypeConfig)
+		if ftc.Name == common.NamespaceResource && ftc.GetTargetType().Kind == common.NamespaceKind &&
+			ftc.GetSourceType().Kind == common.NamespaceKind {
+			nsFTC = ftc
+			break
+		}
+	}
+	if nsFTC == nil {
+		// panic if ftc for namespace does not exist, since no other resource can be synced otherwise
+		klog.Fatal("FederatedTypeConfig for namespaces not found")
+	}
+
+	// ensure the federated namespace since it is a requirement for other controllers
+	if err := c.ensureFederatedObjectCrd(nsFTC); err != nil {
+		klog.Fatalf("Failed to ensure FederatedNamespace CRD: %v", err)
+	}
+
 	c.worker.Run(stopChan)
 
 	// Ensure all goroutines are cleaned up when the stop channel closes
@@ -203,6 +223,7 @@ func (c *Controller) reconcile(qualifiedName common.QualifiedName) worker.Result
 	if c.controllerConfig.CreateCrdForFtcs {
 		if err := c.ensureFederatedObjectCrd(typeConfig); err != nil {
 			klog.Error(fmt.Errorf("cannot ensure federated object CRD for %q: %w", typeConfig.Name, err))
+			return worker.StatusError
 		}
 	}
 
--- a/pkg/controllers/status/controller.go
+++ b/pkg/controllers/status/controller.go
@@ -60,6 +60,8 @@ const (
 // KubeFedStatusController collects the status of resources in member
 // clusters.
 type KubeFedStatusController struct {
+	name string
+
 	// For triggering reconciliation of all target resources. This is
 	// used when a new cluster becomes available.
 	clusterDeliverer *delayingdeliver.DelayingDeliverer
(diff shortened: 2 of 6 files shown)


pkg/controllers/util/federatedinformer.go
```diff
@@ -618,3 +623,54 @@ func (fs *federatedStoreImpl) ClustersSynced(clusters []*fedcorev1a1.FederatedCl
...
+	ctx context.Context,
+	informer FederatedInformer,
+	clusterName string,
+	key string,
+	qualifedName common.QualifiedName,
+	apiResource metav1.APIResource,
+) (*unstructured.Unstructured, bool, error) {
+	if informer.GetTargetStore().ClusterSynced(clusterName) {
+		clusterObj, exists, err := informer.GetTargetStore().GetByKey(clusterName, key)
+		if clusterObj != nil {
```
gary-lgy, line 651: Any chance `clusterObj` will be non-nil garbage when `!exists || err != nil`?

pkg/controllers/util/federatedinformer.go
```diff
@@ -618,3 +623,54 @@ func (fs *federatedStoreImpl) ClustersSynced(clusters []*fedcorev1a1.FederatedCl
...
+}
+
+// GetClusterObject is a helper function to get a cluster object. GetClusterObject first attempts to get the object from
+// the federated informer with the given key. However, if the cache for the cluster is not synced, it will send a GET
+// request to the cluster's apiserver to retrieve the object directly.
+func GetClusterObject(
+	ctx context.Context,
+	informer FederatedInformer,
+	clusterName string,
+	key string,
```
gary-lgy, line 645: Why do we pass both `key` and `qualifiedName`?

pkg/controllers/status/controller.go
```diff
@@ -276,10 +267,6 @@ func (s *KubeFedStatusController) reconcileOnClusterChange() {
 }
 
 func (s *KubeFedStatusController) reconcile(qualifiedName common.QualifiedName) worker.Result {
-	if !s.HasSynced() {
```
gary-lgy, line 279: Do we need to wait for cache sync somewhere?

pkg/controllers/federatedtypeconfig/federatedtypeconfig_controller.go
```diff
@@ -173,6 +173,16 @@ func (c *Controller) Run(stopChan <-chan struct{}) {
 		return
 	}
 
+	if c.controllerConfig.CreateCrdForFtcs {
```
gary-lgy, line 176: We still need to reconcile for each FTC

pkg/controllers/util/federatedinformer.go
limhawjia: Good point. We should perform an error check here.

limhawjia:
I will also move the e2e test fixes into a separate PR

pkg/controllers/federatedtypeconfig/federatedtypeconfig_controller.go
```diff
@@ -173,6 +173,16 @@ func (c *Controller) Run(stopChan <-chan struct{}) {
...
+		ftcs := c.ftcStore.List()
+		for _, obj := range ftcs {
+			ftc := obj.(*fedcorev1a1.FederatedTypeConfig)
+			if err := c.ensureFederatedObjectCrd(ftc); err != nil {
+				klog.Error(fmt.Errorf("cannot ensure federated object CRD for %q: %w", ftc.Name, err))
```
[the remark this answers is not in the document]
limhawjia: My logic here is flawed. I realized I should not have moved the creation of federated CRDs out of the reconcile function because:

1. We need to handle the creation of new FTCs after the controller-manager has started
2. The error returned might be recoverable (such as timeout when getting from apiserver) and we can use the workerqueue semantics to retry

However, I think panicking is not a good idea because even if we fail to create the federated CRD for a single FTC, there are other controllers and ftcs that can continue to work properly and we should allow them to run.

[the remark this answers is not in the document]
limhawjia: My initial motivation for moving it out of the reconcile function is because of a potential deadlock caused by:

1. Sync controller's dependency on the federatednamespace crd
2. FTC controller starting subcontrollers synchronously

So what might happen:

1. FTC controller receives events for a FTC that are not the namespace FTC
2. All workers are used to process these events
3. Each worker attempts to start the sync controller synchronously
4. Each sync controller waits for the federated namespace cache to sync (but this can never happen because the federated namespace CRD was never created)
5. All workers are blocked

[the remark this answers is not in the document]
limhawjia: The new approach is to only ensure the federated crd for namespaces outside the reconcile function and panic if we fail to do so. This is because the federated namespace is a requirement that will block some of the controllers - namely sync and nsautopropagation.

The federated crd for other resources will continue to be handled in the reconcile function.

cc @gary-lgy @mrlihanbo

gary-lgy (approved):

PR #86, opened 2023-05-31
ci: add docker build action

SOF3:


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,46 @@
...
+          else
+            echo "tag=dev"
+          fi >>$GITHUB_OUTPUT
+
+      - name: Build docker image
+        env:
+          REGISTRY: ghcr.io/kubewharf
+          TAG: ${{ steps.tag.outputs.tag }}
+          ARCHS: amd64
+          OUTPUT_TYPE: registry
```
qclc, line 42: Use the `registry` OUTPUT_TYPE, which is shorthand for `type=image,push=true`. But there is also github action that perform push below, and it seems that two push operations have been performed.

hack/make-rules/build-images.sh
```diff
@@ -34,7 +34,7 @@ platform_array=("${arch_array[@]/#/linux/}")
 PLATFORMS=$(IFS=","; echo "${platform_array[*]}")
 REGISTRY=${REGISTRY:-"ghcr.io/kubewharf"}
 OUTPUT_TYPE="docker"
```
qclc, line 36:
```suggestion
OUTPUT_TYPE=${OUTPUT_TYPE:-"docker"}
```
The value of `OUTPUT_TYPE` here is fixed as the `docker` type. If we need to support multiple types of `OUTPUT_TYPE`, please change it to receive a different variable value.

.github/workflows/docker-build.yml
SOF3: According to my previous test, `registry` only tags the image without pushing.

SOF3: Oh it might be related to the comment below.

qclc (commented):
lgtm

gary-lgy (approved):

PR #88, opened 2023-06-01
fix:record labels and annotations after apply overrider to resource

wy-lucky:
fix:record labels and annotations after apply overrider to resource

### DIFF
--- a/pkg/controllers/sync/dispatch/managed.go
+++ b/pkg/controllers/sync/dispatch/managed.go
@@ -342,6 +342,8 @@ func (d *managedDispatcherImpl) Create(clusterName string) {
 			return d.recordOperationError(fedtypesv1a1.ApplyOverridesFailed, clusterName, op, err)
 		}
 
+		recordPropagatedLabelsAndAnnotations(obj)
+
 		ctx, cancel := context.WithTimeout(context.Background(), d.dispatcher.timeout)
 		defer cancel()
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/sync/dispatch/managed.go
```diff
@@ -342,6 +342,11 @@ func (d *managedDispatcherImpl) Create(clusterName string) {
 			return d.recordOperationError(fedtypesv1a1.ApplyOverridesFailed, clusterName, op, err)
 		}
 
+		err = AddAnnotationToRecordLabelsAndAnnotations(obj)
```
gary-lgy, line 345:
```suggestion
		err = RecordPropagatedLabelsAndAnnotations(obj)
```

pkg/controllers/sync/dispatch/retain.go
```diff
@@ -96,6 +96,25 @@ func RetainOrMergeClusterFields(
 	return nil
 }
 
+func AddAnnotationToRecordLabelsAndAnnotations(obj *unstructured.Unstructured) error {
+	// Record the annotation/label keys in the template so we can diff it against the cluster object during retention
+	// to determine whether an annotation/label has been deleted from the template.
+	if _, err := annotationutil.AddAnnotation(
```
gary-lgy, line 102: We can avoid having to return an error by using `obj.GetAnnotations` and `obj.SetAnnotations` directly.

pkg/controllers/sync/resource.go
```diff
@@ -197,21 +197,6 @@ func (r *federatedResource) ObjectForCluster(clusterName string) (*unstructured.
 		obj.SetFinalizers(nil)
 	}
 
-	// Record the annotation/label keys in the template so we can diff it against the cluster object during retention
-	// to determine whether an annotation/label has been deleted from the template.
-	if _, err := annotationutil.AddAnnotation(
-		obj, common.TemplateAnnotationKeys,
```
gary-lgy, line 203: Let's rename `TemplateAnnotationKeys` to `PropagatedAnnotationKeys` since it's not just the template we're recording anymore.

gary-lgy (approved):
In order to record only the propagated keys, we need to record the keys before merging with cluster object. However, the recorded keys must include keys added from overrides.

As a result, we have to follow this sequence `override` -> `record` -> `merge`.

Overrides used to be applied after merge to give overrides a chance to overwrite any merged fields (ref: https://github.com/kubernetes-retired/kubefed/blob/master/docs/userguide.md#overriding-retained-fields). We find this use case unreasonable since it overloads the responsibility of overrides, which should only be to modify the desired object template for each cluster. The desired object template, as usual, does not take precedence over any merging logic. Thus we intentionally change the old behaviour to apply overrides before merging with cluster object.

PR #96, opened 2023-06-08
fix: automigration

limhawjia:
### The last implementation of the automigration controller breaks when overlapping label selectors are used.

**Consider the following case:**

1. There are 2 deployments, `dp-a` and `dp-b` with overlapping label selectors.
2. Estimated capacity is computed by taking `totalReplicas - numUnschedulablePods`.
3. Any unschedulable pods from one of the deployments will be wrongly counted when computing the estimated capacity of the other deployment.
4. This causes estimated capacity to be underestimated, or even become negative

**When estimated capacity is underestimated:**

1. Automigration will unnecessarily migrate pods, affecting stability.

**When estimated capacity is negative:**
1. The planner's logic does not even work, and scheduling becomes unpredictable.
2. It could lead to an infinite loop where deployments are scaled up infinitely.
    a. The planner creates new pods to "migrate" the "extra" unschedulable pods for `dp-a`.
    b. If these new pods are also unschedulable, it would cause the estimated capacity of `dp-b` to decrease.
    c. When this happens, the planner will repeat step 2a for `dp-b`, potentially resulting in an infinite loop

### The old algorithm for estimating capacity considers pods from both old and new replicasets

We should only consider pods from the latest replicaset when calculating `estimatedCapacty`. This is because only the latest replicaset will match the deployment's pod spec template, and different pod spec templates have different schedulability.

As a result, considering pods from old replicasets will cause us to either overestimate or underestimate the capacity of a deployment.

Consider the following example:

A deployment is in the middle of a rolling update. Assume that the old replicaset is inherently unschedulable (perhaps due to a misconfigured nodeSelector) while all pods of the new replicaset are inherently schedulable + the cluster has enough resources for them. The deployment is in the state:

```
Desired replicas = 5
Old replicaset = 0 schedulable, 5 unschedulable pods
New replicaset = 2 schedulable, 0 unschedulable pods
```
If we consider pods from both replicasets and compute `estimatedCapacity = totalReplicas - unschdulablePods`, we would get `estimatedCapacity = 7 - 5 = 2`. This would cause the planner to unnecessarily migrate 3 pods that would otherwise have been schedulable.

This is a case of underestimating capacity.

Consider another example:

We have a deployment in the middle of a rolling update. The deployment is in the state:

```
Desired replicas = 10
Old replicaset = 15 schedulable, 0 unschedulable pods
New replicaset = 1 schedulable, 4 unschedulable pods
```

If we consider pods from both replicasets and compute `estimatedCapacty = totalReplicas - unschedulablePods`, we would get `estimatedCapacity = 20 - 4 = 16`. This would cause the planner to not migrate any pods even though we have 4 unschedulable.

This is a case of overestimating capacity.

### Key changes:

**Estimate capacity algorithm**

The estimate capacity algorithm is changed to:

If `totalNewReplicas >= desiredReplicas`, `estimatedCapacity = schedulableNewPods`, else `estimatedCapacity = desiredReplicas - unschedulableNewPods`.

The reason for handling the 2 cases differently is that when `totalNewReplicas >= desiredReplicas`, we can immediately get the capacity by counting the number of schedulable pods. However, when `totalNewReplicas < desiredReplicas`, we have uncreated pods and will have to consider them to be schedulable to prevent them from being migrated away before they are even created.

**Automigration controller**

1. Sets a lower bound for estimated capacity to be 0.
3. Introduce a plugin mechanism to resolve pods for a workload based on the target GVR.
4.  Add a plugin for deployments to accurately retrieve pods from the latest replicaset.

**Automigration e2e tests**

1. Added 2 new e2e tests to test automigration in cases when overlapping label selectors are used for deployments.

**Planner**

1. Add a safeguard in planner to ignore negative estimated capacity values


pkg/controllers/automigration/controller.go
```diff
@@ -327,14 +338,20 @@ func (c *Controller) estimateCapacity(
 
 		unschedulable, nextCrossIn := countUnschedulablePods(pods, time.Now(), unschedulableThreshold)
 		keyedLogger.V(3).Info("Analyzed pods",
```
gary-lgy, line 340: Should we print this at V(1) if there are unschedulable pods?

pkg/controllers/automigration/controller.go
```diff
@@ -301,24 +301,35 @@ func (c *Controller) estimateCapacity(
...
+		unsClusterObj := clusterObj.Object.(*unstructured.Unstructured)
 
 		// NOTE: these should follow Deployment's status semantics and should not include terminating pods
-		totalReplicas, readyReplicas, err := c.getTotalAndReadyReplicas(unsObj)
+		//
+		// TODO: what if some workloads do not have a field that represents the totalReplicas or readyReplicas?
+		totalReplicas, readyReplicas, err := c.getTotalAndReadyReplicas(unsClusterObj)
 		if err != nil {
-			keyedLogger.Error(err, "Failed to get total and ready replicas from object", "cluster", clusterObj.ClusterName)
+			// TODO: since we are now using desiredReplicas - unschedulablePods to calculate estimatedCapacity,
```
gary-lgy, line 313: Why is this a TODO?

pkg/controllers/automigration/controller.go
```diff
@@ -327,14 +338,20 @@ func (c *Controller) estimateCapacity(
...
-			"cluster", clusterObj.ClusterName,
+			"desired", desiredReplicas,
 			"total", totalReplicas,
 			"ready", readyReplicas,
 			"unschedulable", unschedulable,
 		)
+
 		if unschedulable > 0 {
-			estimatedCapacity[clusterObj.ClusterName] = totalReplicas - int64(unschedulable)
+			if int64(unschedulable) > desiredReplicas {
```
gary-lgy, line 348: Should we add a comment here?

pkg/controllers/automigration/controller.go
```diff
@@ -301,24 +301,35 @@ func (c *Controller) estimateCapacity(
...
 			continue
 		}
-
 		if totalReplicas == readyReplicas {
-			// no unschedulable pods
+			// no unschedulable pods, avoid unnecessary pod listing
+			continue
+		}
+
+		desiredReplicas, err := c.getDesiredReplicas(unsClusterObj)
```
gary-lgy, line 323: Should we add a comment to explain why we should use spec.replicas instead of status.replicas?

pkg/controllers/automigration/controller.go
limhawjia: V(2) seems more appropriate

pkg/controllers/automigration/controller.go
limhawjia: It's related to the TODO above. Will merge them together.

limhawjia:
@gary-lgy the estimate capacity algorithm underwent some more changes, ptal

pkg/controllers/automigration/controller.go
```diff
@@ -301,40 +301,64 @@ func (c *Controller) estimateCapacity(
...
-		// NOTE: these should follow Deployment's status semantics and should not include terminating pods
-		totalReplicas, readyReplicas, err := c.getTotalAndReadyReplicas(unsObj)
-		if err != nil {
-			keyedLogger.Error(err, "Failed to get total and ready replicas from object", "cluster", clusterObj.ClusterName)
+		unsClusterObj := clusterObj.Object.(*unstructured.Unstructured)
+
+		// This is an optimization to skip pod listing when there are no unschedulable pods.
+		totalReplicas, readyReplicas, err := c.getTotalAndReadyReplicas(unsClusterObj)
+		if err == nil && totalReplicas == readyReplicas {
+			keyedLogger.V(3).Info("No unschedulable pods found, skip estimate capacity")
```
gary-lgy, line 311:
```suggestion
			keyedLogger.V(3).Info("No unschedulable pods found, skip estimating capacity")
```

pkg/controllers/automigration/controller.go
```diff
@@ -301,40 +301,64 @@ func (c *Controller) estimateCapacity(
 	estimatedCapacity := make(map[string]int64, len(clusterObjs))
 
 	for _, clusterObj := range clusterObjs {
-		unsObj := clusterObj.Object.(*unstructured.Unstructured)
+		keyedLogger := keyedLogger.WithValues("cluster", clusterObj.ClusterName)
```
gary-lgy, line 304: Do we want to insert the new logger into the context? We should probably provide a helper for this in case we forget.

pkg/controllers/automigration/controller.go
```diff
@@ -377,43 +401,38 @@ func (c *Controller) getTotalAndReadyReplicas(
...
+	plugin, err := plugins.ResolvePlugin(c.typeConfig)
 	if err != nil {
-		return nil, false, fmt.Errorf("failed to convert label selector to selector: %w", err)
+		return nil, false, fmt.Errorf("failed to get plugin for FTC: %w", err)
 	}
 
 	client, err := c.federatedInformer.GetClientForCluster(clusterName)
 	if err != nil {
-		return nil, false, fmt.Errorf("failed to get client for cluster: %w", err)
+		return nil, true, fmt.Errorf("failed to get client for cluster: %w", err)
```
gary-lgy, line 427: Why do we require backoff here?

pkg/lifted/kubernetes/pkg/util/hash/hash_test.go
```diff
@@ -0,0 +1,159 @@
...
+This file may have been modified by The KubeAdmiral Authors
+("KubeAdmiral Modifications"). All KubeAdmiral Modifications
+are Copyright 2023 The KubeAdmiral Authors.
+*/
+
+/*
+This file is lifted from k8s.io/kubernetes/pkg/util/hash/hash.go
+*/
+
+package hash
```
gary-lgy, line 25: Do we need nolint here

pkg/controllers/automigration/controller.go
limhawjia: If we fail to get the client for cluster, it could mean that the federatedinformer is out of date. In this case, I think should retry until the federatedinformer becomes up to date?

pkg/controllers/automigration/plugins/deployments.go
```diff
@@ -0,0 +1,112 @@
...
+			ret = append(ret, &rsList.Items[i])
+		}
+		return ret, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	newRS := deploymentutil.FindNewReplicaSet(deployment, rsList)
+	podList, err := deploymentutil.ListPods(
```
gary-lgy, line 66: Should we handle newRs==nil?

pkg/controllers/util/planner/planner.go
```diff
@@ -218,7 +218,7 @@ func getDesiredPlan(
 	// Assign each cluster the minimum number of replicas it requested.
 	for _, preference := range preferences {
 		min := minInt64(preference.MinReplicas, remainingReplicas)
-		if capacity, hasCapacity := estimatedCapacity[preference.clusterName]; hasCapacity && capacity < min {
+		if capacity, hasCapacity := estimatedCapacity[preference.clusterName]; hasCapacity && capacity >= 0 && capacity < min {
```
gary-lgy, line 221: estimatedCapacity is used in a few other places too. Should we also update those?

limhawjia: yes, then i think it will be better to do a preprocessing step where we remove all negative estimatedCapacity values

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +238,343 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+		})
+
+		// This function creates an override policy that results in only a single pod of a deployment being schedulable
+		// in a given cluster using a combination of node affinity and pod anti-affinity.
+		getOverridePolicy := func(cluster *fedcorev1a1.FederatedCluster, deploy *appsv1.Deployment) *fedcorev1a1.OverridePolicy {
+			var selectedNode string
+			var selectedLabelKey, selectedLabelValue string
+
+			for k, v := range dp1.Spec.Template.GetLabels() {
+				selectedLabelKey = k
```
gary-lgy, line 291: How do we know there are labels for sure?

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +238,343 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+							Path:     "/spec/template/spec/affinity",
+							Value: apiextensionsv1.JSON{
+								Raw: []byte(
+									//nolint:lll
+									`{"nodeAffinity": {"requiredDuringSchedulingIgnoredDuringExecution": {"nodeSelectorTerms": []}}, "podAntiAffinity": {"requiredDuringSchedulingIgnoredDuringExecution": []}}`,
+								),
+							},
+						},
+						{
+							Path: "/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms",
```
gary-lgy, line 316: With PodAntiAffinity, is NodeAffinity necessary?

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +238,343 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+		ginkgo.By("Creating PropagationPolicy", func() {
+			propPolicy, err = f.HostFedClient().CoreV1alpha1().PropagationPolicies(f.TestNamespace().Name).Create(
+				ctx,
+				propPolicy,
+				metav1.CreateOptions{},
+			)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred(), framework.MessageUnexpectedError)
+		})
+
+		// This function creates an override policy that results in only a single pod of a deployment being schedulable
```
gary-lgy, line 284: Can't we create dp2 directly in one of the member clusters, and test whether dp1 is migrated unnecessarily?

pkg/controllers/automigration/plugins/deployments.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return nil, err
+	}
+	fieldSelector, err := fields.ParseSelector(opts.FieldSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	return &client.ListOptions{
+		LabelSelector: labelSelector,
```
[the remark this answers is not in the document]
limhawjia: this made me realized that `client.ListOptions` has a `Raw` field which allows me to skip parsing the selectors

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +238,343 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+					g.Expect(clusterDp1.Spec.Replicas).To(gomega.HaveValue(gomega.Equal(replicasPerCluster)))
+
+					clusterDp2, err := f.ClusterKubeClient(ctx, c).AppsV1().Deployments(f.TestNamespace().Name).Get(
+						ctx, dp2.Name, metav1.GetOptions{},
+					)
+					if err != nil {
+						gomega.Expect(err).To(gomega.Satisfy(apierrors.IsNotFound))
+					}
+					g.Expect(err).NotTo(gomega.HaveOccurred())
+					g.Expect(clusterDp2.Spec.Replicas).To(gomega.HaveValue(gomega.Equal(replicasPerCluster)))
```
gary-lgy, line 555: Should we also assert readyReplicas=0 just in case?

pkg/controllers/automigration/controller.go
gary-lgy: This is handled in other controllers by reconciling everything on `Cluster(A|Un)available`, which is wasteful and dangerous. Let's return true here and revisit this after we replace `FederatedInformer` with `FederatedInformerManager`

pkg/controllers/util/planner/planner.go
```diff
@@ -215,10 +215,18 @@ func getDesiredPlan(
 	plan := make(map[string]int64, len(preferences))
 	overflow := make(map[string]int64, len(preferences))
 
+	// This algorithm does not support negative estimatedCapacity values.
+	santizedEstimatedCapacity := map[string]int64{}
```
gary-lgy, line 219: This probably needs to be at the top of `Plan`, or outside it.

pkg/controllers/automigration/plugins/deployments.go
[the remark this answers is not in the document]
gary-lgy: `Raw` is not fully respected. Pls make sure it's safe to use `Raw` alone.

pkg/controllers/util/planner/planner.go
limhawjia: agreed. will move it

pkg/controllers/automigration/plugins/deployments.go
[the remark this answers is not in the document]
limhawjia: Verified that it is respected for our client implementation

test/e2e/automigration/automigration.go
limhawjia: because we have labels in the deployment template that we generate

test/e2e/automigration/automigration.go
limhawjia: yes, without nodeaffinity all the nodes will need to have the same topology key

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +239,353 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+
+		expectedDistributionDp1 := map[string]int32{}
+		for _, cluster := range clusters {
+			if cluster.Name == clusterToMigrateFromForDp1.Name {
+				continue
+			}
+
+			// The deployment in clusterToMigrateFromForDp1 will only have 1 schedulable pod. The remaining should be
+			// migrated to the remaining clusters evenly (based on the propagation policy).
+			expectedDistributionDp1[cluster.Name] = replicasPerCluster + ((replicasPerCluster)-1)/int32(len(clusters)-1)
```
gary-lgy, line 423: This division only works when there is no remainder (such as this case). Can we move this closer to where `replicasPerCluster` is defined so we can change them in tandem if required.

test/e2e/automigration/automigration.go
```diff
@@ -228,4 +239,353 @@ var _ = ginkgo.Describe("auto migration", autoMigrationTestLabel, func() {
...
+					ctx, dp2.Name, metav1.GetOptions{},
+				)
+				gomega.Expect(err).ToNot(gomega.HaveOccurred())
+				g.Expect(clusterDp.Spec.Replicas).To(gomega.HaveValue(gomega.Equal(replicasPerCluster)))
+			}).WithTimeout(5 * time.Second).Should(gomega.Succeed())
+		}
+
+		expectedDistributionDp1 := map[string]int32{}
+		for _, cluster := range clusters {
+			if cluster.Name == clusterToMigrateFromForDp1.Name {
```
gary-lgy, line 417: Why don't we write `expectedDistributionDp1[cluster.Name] = replicasPerCluster`?

gary-lgy (approved):

PR #101, opened 2023-06-12
chore: add crd ftc to propagate crd

JackZxj:


gary-lgy:
@limhawjia Should this be in `extra`? Would enabling this by default bring any problems (I can't seem to think of any)?

limhawjia:
> @limhawjia Should this be in `extra`? Would enabling this by default bring any problems (I can't seem to think of any)?

i think enabling it by default is fine

JackZxj:
Has been moved to the default, take a look plz~ @gary-lgy @limhawjia

limhawjia (approved):

PR #110, opened 2023-06-13
feat: add status aggregator plugins for job and pod

JackZxj:
When the FTC of the pod and job enables `statusAggregation`, now we can get the aggregated status of job and pod on the native resources on the federation side.

For pods, InitContainerStatuses and ContainerStatuses are collected. The aggregated startTime depends on the earliest startTime in the member cluster pod, and the aggregated pod phase is generated by the following priorities:
1. If there is any failed pod, the aggregation phase is `Failed`
2. If there is any pending pod, or a pod whose status is empty, the aggregation phase is `Pending`
3. If there is any running pod, the aggregation phase is `Running`
4. If **all** pods are succeeded, the aggregation phase is `Succeeded`

For jobs, the sum of all Active, Succeeded and Failed pods will be collected. The aggregated startTime depends on the earliest startTime in the member cluster job. The aggregated job Condition is generated by the following priority:
1. If **all** the jobs are `Completed` or `Failed`, and there is any failed job, the aggregation status is `Failed`, and the completionTime is the time of the last completed job
2. If **all** the jobs are `Completed`, the aggregation status is `Completed`, and the completionTime is the time of the last completed job
3. Otherwise, the job is still in processing

Also, it shows some extra messages about how the job/pod works in the member cluster, which I think might be useful for developers or debuggers.

Here is an example result:

`kubectl get pod test-po -oyaml`
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubeadmiral.io/scheduling: '{"generation":null,"fedGeneration":2,"placement":["kubeadmiral-member-1","kubeadmiral-member-2"]}'
    kubeadmiral.io/status: '{"clusters":[{"name":"kubeadmiral-member-1","generation":0},{"name":"kubeadmiral-member-2","generation":0}]}'
    kubeadmiral.io/syncing: '{"generation":null,"fedGeneration":2,"clusters":[{"name":"kubeadmiral-member-1","status":"OK"},{"name":"kubeadmiral-member-2","status":"OK"}]}'
  creationTimestamp: "2023-06-12T08:23:07Z"
  finalizers:
  - kubeadmiral.io/federate-controller
  labels:
    kubeadmiral.io/propagation-policy-name: pp-2cluster-duplicate
    run: test-po
  name: test-po
  namespace: default
  resourceVersion: "1349830"
  uid: 31348c96-0bf1-4cfa-8418-251616f55d2c
spec:
  automountServiceAccountToken: false
  containers:
  - image: nginx:alpine
    imagePullPolicy: IfNotPresent
    name: test-po
...
status:
  containerStatuses:
  - containerID: containerd://26bf95ca6d04e9c5d15cea130418f96740217db784898adeabb907e892e8c3a5
    image: docker.io/library/nginx:alpine
    imageID: docker.io/library/nginx@sha256:2e776a66a3556f001aba13431b26e448fe8acba277bf93d2ab1a785571a46d90
    lastState: {}
    name: test-po (kubeadmiral-member-2)
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2023-06-12T13:23:19Z"
  message: Pod pending in clusters [kubeadmiral-member-1], and running in clusters
    [kubeadmiral-member-2]
  phase: Pending
  startTime: "2023-06-12T13:23:18Z"
```

`kubectl get job test-job -oyaml`

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    kubeadmiral.io/scheduling: '{"generation":null,"fedGeneration":3,"placement":["kubeadmiral-member-1","kubeadmiral-member-2"]}'
    kubeadmiral.io/status: '{"clusters":[{"name":"kubeadmiral-member-1","generation":0},{"name":"kubeadmiral-member-2","generation":0}]}'
    kubeadmiral.io/syncing: '{"generation":null,"fedGeneration":3,"clusters":[{"name":"kubeadmiral-member-1","status":"OK"},{"name":"kubeadmiral-member-2","status":"OK"}]}'
  creationTimestamp: "2023-06-12T14:01:01Z"
  finalizers:
  - kubeadmiral.io/federate-controller
  labels:
    app: test-job
    job-name: test-job
    kubeadmiral.io/override-policy-name: job-failed
    kubeadmiral.io/propagation-policy-name: pp-2cluster-duplicate
  name: test-job
  namespace: default
  resourceVersion: "1353660"
  uid: f2e7a03e-10d6-44c3-88b9-9f5c2075c796
spec:
  activeDeadlineSeconds: 300
  backoffLimit: 6
  completions: 3
  parallelism: 2
  selector:
    matchLabels:
      controller-uid: f2e7a03e-10d6-44c3-88b9-9f5c2075c796
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: test-job
        controller-uid: f2e7a03e-10d6-44c3-88b9-9f5c2075c796
        job-name: test-job
    spec:
      containers:
      - args:
        - -c
        - date; echo running; sleep 60; date; echo done
        command:
        - /bin/sh
        image: nginx:alpine
        imagePullPolicy: IfNotPresent
...
status:
  completionTime: "2023-06-12T14:05:20Z"
  conditions:
  - lastProbeTime: "2023-06-12T14:05:30Z"
    lastTransitionTime: "2023-06-12T14:05:30Z"
    message: Job completed in clusters [kubeadmiral-member-2] and failed in member
      clusters [kubeadmiral-member-1]
    reason: Mixed
    status: "True"
    type: Failed
  failed: 8
  startTime: "2023-06-12T14:03:18Z"
  succeeded: 3
```

### DIFF
--- a/pkg/controllers/statusaggregator/plugins/job.go
+++ b/pkg/controllers/statusaggregator/plugins/job.go
@@ -1,0 +1,163 @@
+/*
+Copyright 2023 The KubeAdmiral Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugins
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+	"time"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/common"
+)
+
+type JobPlugin struct{}
+
+func NewJobPlugin() *JobPlugin {
+	return &JobPlugin{}
+}
+
+func (receiver *JobPlugin) AggregateStatuses(
+	ctx context.Context,
+	sourceObject, fedObject *unstructured.Unstructured,
+	clusterObjs map[string]interface{},
+	clusterObjsUpToDate bool,
+) (*unstructured.Unstructured, bool, error) {
+	logger := klog.FromContext(ctx).WithValues("status-aggregator-plugin", "jobs")
+
+	aggregatedStatus := &batchv1.JobStatus{}
+	var completionTime *metav1.Time
+	var finishedCount int
+	var completedJobs, failedJobs []string
+
+	for clusterName, clusterObj := range clusterObjs {
+		logger := klog.FromContext(ctx).WithValues("cluster-name", clusterName)
+
+		utd := clusterObj.(*unstructured.Unstructured)
+		// For status of job
+		var found bool
+		status, found, err := unstructured.NestedMap(utd.Object, common.StatusField)
+		if err != nil || !found {
+			logger.Error(err, "Failed to get status of cluster object")
+			return nil, false, err
+		}
+
+		if status == nil {
+			continue
+		}
+
+		jobStatus := &batchv1.JobStatus{}
+		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(status, jobStatus); err != nil {
+			logger.Error(err, "Failed to convert the status of cluster object")
+			return nil, false, err
+		}
+
+		if aggregatedStatus.StartTime.IsZero() || jobStatus.StartTime.Before(aggregatedStatus.StartTime) {
+			aggregatedStatus.StartTime = jobStatus.StartTime
+		}
+		if !jobStatus.CompletionTime.IsZero() {
+			finishedCount++
+			completedJobs = append(completedJobs, clusterName)
+			if completionTime.IsZero() || completionTime.Before(jobStatus.CompletionTime) {
+				completionTime = jobStatus.CompletionTime
+			}
+		} else if IsJobFinishedWithFailed(jobStatus) {
+			finishedCount++
+			failedJobs = append(failedJobs, clusterName)
+		}
+
+		aggregatedStatus.Active += jobStatus.Active
+		aggregatedStatus.Succeeded += jobStatus.Succeeded
+		aggregatedStatus.Failed += jobStatus.Failed
+	}
+	if finishedCount > 0 && finishedCount == len(clusterObjs) {
+		now := time.Now()
+		var conditionType batchv1.JobConditionType
+		var reason, message string
+		sort.Strings(completedJobs)
+		sort.Strings(failedJobs)
+
+		switch {
+		case len(completedJobs) > 0 && len(failedJobs) > 0:
+			conditionType = batchv1.JobFailed
+			reason = "Mixed"
+			message = fmt.Sprintf("Job completed in clusters [%s] and failed in member clusters [%s]",
+				strings.Join(completedJobs, ","), strings.Join(failedJobs, ","))
+		case len(completedJobs) > 0:
+			conditionType = batchv1.JobComplete
+			reason = "Completed"
+			message = fmt.Sprintf("Job completed in clusters [%s]", strings.Join(completedJobs, ","))
+			aggregatedStatus.CompletionTime = completionTime
+		default:
+			conditionType = batchv1.JobFailed
+			reason = "Failed"
+			message = fmt.Sprintf("Job failed in clusters [%s]", strings.Join(failedJobs, ","))
+		}
+
+		aggregatedStatus.Conditions = append(aggregatedStatus.Conditions, batchv1.JobCondition{
+			Type:               conditionType,
+			Status:             corev1.ConditionTrue,
+			LastProbeTime:      metav1.NewTime(now),
+			LastTransitionTime: metav1.NewTime(now),
+			Reason:             reason,
+			Message:            message,
+		})
+	}
+
+	newStatus, err := runtime.DefaultUnstructuredConverter.ToUnstructured(aggregatedStatus)
+	if err != nil {
+		logger.Error(err, "Failed to convert aggregated status to unstructured")
+		return nil, false, err
+	}
+
+	oldStatus, _, err := unstructured.NestedMap(sourceObject.Object, common.StatusField)
+	if err != nil {
+		logger.Error(err, "Failed to get old status of source object")
+		return nil, false, err
+	}
+
+	// update status of source object if needed
+	needUpdate := false
+	if !reflect.DeepEqual(newStatus, oldStatus) {
+		if err := unstructured.SetNestedMap(sourceObject.Object, newStatus, common.StatusField); err != nil {
+			logger.Error(err, "Failed to set the new status on source object")
+			return nil, false, err
+		}
+		needUpdate = true
+	}
+
+	return sourceObject, needUpdate, nil
+}
+
+// IsJobFinishedWithFailed checks whether the given Job has finished execution with failed condition.
+func IsJobFinishedWithFailed(jobStatus *batchv1.JobStatus) bool {
+	for _, c := range jobStatus.Conditions {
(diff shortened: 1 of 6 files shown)


pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+	sourceObject, fedObject *unstructured.Unstructured,
+	clusterObjs map[string]interface{},
+	clusterObjsUpToDate bool,
+) (*unstructured.Unstructured, bool, error) {
+	logger := klog.FromContext(ctx).WithValues("status-aggregator-plugin", "jobs")
+
+	needUpdate := false
+	aggregatedStatus := &batchv1.JobStatus{}
+	var completionTime *metav1.Time
+	var completionCount int
```
gary-lgy, line 54: `completionCount` might not be the most appropriate name since it counts both `completed` and `failed` jobs. What about `finishedCount`?

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+			conditionType = batchv1.JobComplete
+			reason = "Completed"
+			message = fmt.Sprintf("Job completed in clusters [%s]", strings.Join(jobCompleted, ","))
+		default:
+			conditionType = batchv1.JobFailed
+			reason = "Failed"
+			message = fmt.Sprintf("Job failed in clusters [%s]", strings.Join(jobFailed, ","))
+		}
+
+		aggregatedStatus.CompletionTime = completionTime
```
gary-lgy, line 120: When there are failed jobs, should we set `CompletionTime`? https://github.com/kubernetes/kubernetes/blob/a17149e1a189050796ced469dbd78d380f2ed5ef/pkg/controller/job/job_controller.go#L568 The Job controller seems to only set `CompletionTime` when the Job succeeded.

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+		aggregatedStatus.Conditions = append(aggregatedStatus.Conditions, batchv1.JobCondition{
```
gary-lgy, line 121: Would this append the same `Failed` or `Completed` conditions infinitely during successive reconciles? We should either overwrite the old conditions here, or prevent updating the status if the job is already finished. @mrlihanbo @limhawjia Which one sounds more reasonable?

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (receiver *JobPlugin) AggregateStatuses(
+	ctx context.Context,
+	sourceObject, fedObject *unstructured.Unstructured,
+	clusterObjs map[string]interface{},
+	clusterObjsUpToDate bool,
+) (*unstructured.Unstructured, bool, error) {
+	logger := klog.FromContext(ctx).WithValues("status-aggregator-plugin", "jobs")
+
+	needUpdate := false
```
gary-lgy, line 51: Let's move `needUpdate` closer to where it is used.

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+	aggregatedStatus := &batchv1.JobStatus{}
+	var completionTime *metav1.Time
+	var completionCount int
+	var jobCompleted, jobFailed []string
```
gary-lgy, line 55:
```suggestion
	var completedJobs, failedJobs []string
```

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+		} else if IsJobFinishedWithFailed(jobStatus) {
+			completionCount++
+			jobFailed = append(jobFailed, clusterName)
+		}
+
+		aggregatedStatus.Active += jobStatus.Active
+		aggregatedStatus.Succeeded += jobStatus.Succeeded
+		aggregatedStatus.Failed += jobStatus.Failed
+	}
+	if completionCount > 0 && completionCount == len(clusterObjs) {
```
gary-lgy, line 97: If there is a `Failed` job, should we wait until all jobs are finished before marking the host job as `Failed`? @mrlihanbo @limhawjia

pkg/controllers/statusaggregator/plugins/job.go
limhawjia: I think we should try to follow the semantics of the job controller as much as possible, which seems to be to skip updating finished jobs

pkg/controllers/statusaggregator/plugins/job.go
limhawjia: Semantically, I think a single failed job should mean that the host job has failed. However, setting it immediately would cause us to lose information about the remaining jobs if we choose not to update finished jobs (as described below)... I'm still not sure what's the best approach

pkg/controllers/statusaggregator/plugins/job.go
limhawjia: After giving it some thought, we should continue to update it so as to not lose information about the remaining clusters' jobs after the host job is marked as `Failed`. 

@gary-lgy there shouldn't be an infinite loop as the FederatedJobStatus itself should not be a reconcile trigger?

limhawjia: Can we solve the infinite reconcile loop problem by checking the type, status, reason and message? If they are all equal, we skip updating.

pkg/controllers/statusaggregator/plugins/job.go
limhawjia: We should follow the semantics of the Job controller and only set `CompletionTime` when the job succeeds

pkg/controllers/statusaggregator/plugins/job.go
```diff
@@ -0,0 +1,163 @@
...
+
+	for clusterName, clusterObj := range clusterObjs {
+		logger := klog.FromContext(ctx).WithValues("cluster-name", clusterName)
+
+		utd := clusterObj.(*unstructured.Unstructured)
+		// For status of job
+		var found bool
+		status, found, err := unstructured.NestedMap(utd.Object, common.StatusField)
+		if err != nil || !found {
+			logger.Error(err, "Failed to get status of cluster object")
```
limhawjia, line 65: If status is not found, perhaps we should continue with status aggregation?

pkg/controllers/statusaggregator/plugins/pod.go
```diff
@@ -0,0 +1,153 @@
...
+
+	for clusterName, clusterObj := range clusterObjs {
+		logger := klog.FromContext(ctx).WithValues("cluster-name", clusterName)
+
+		utd := clusterObj.(*unstructured.Unstructured)
+		// For status of pod
+		var found bool
+		status, found, err := unstructured.NestedMap(utd.Object, common.StatusField)
+		if err != nil || !found {
+			logger.Error(err, "Failed to get status of cluster object")
```
limhawjia, line 67: We can continue aggregating the statuses of other cluster pods even if status is not found for one of them

pkg/controllers/statusaggregator/plugins/pod.go
```diff
@@ -0,0 +1,153 @@
...
+			aggregatedStatus.Phase = phase
+		}
+		sort.Strings(phases[phase])
+		if aggregatedStatus.Message == "" {
+			aggregatedStatus.Message = fmt.Sprintf("Pod %s in clusters [%s]",
+				strings.ToLower(string(phase)), strings.Join(phases[phase], ","))
+			continue
+		}
+		aggregatedStatus.Message = fmt.Sprintf("%s, and %s in clusters [%s]",
+			aggregatedStatus.Message, strings.ToLower(string(phase)), strings.Join(phases[phase], ","))
```
limhawjia, line 116: are we missing a `break` here?

pkg/controllers/statusaggregator/plugins/pod.go
```diff
@@ -0,0 +1,153 @@
...
+		// For status of pod
+		var found bool
+		status, found, err := unstructured.NestedMap(utd.Object, common.StatusField)
+		if err != nil || !found {
+			logger.Error(err, "Failed to get status of cluster object")
+			return nil, false, err
+		}
+
+		if status == nil {
+			continue
```
limhawjia, line 72: if status is nil, should we set the phase to pending by default? with the current logic, the pod will simply not be considered in the aggregation below and we might set the host pod's phase to `Succeeded`

pkg/controllers/statusaggregator/plugins/pod.go
limhawjia: oh i misinterpreted... i thought that we were replacing the  phase and message with each loop. this is actually ok

limhawjia (approved):

PR #112, opened 2023-06-13
fix: the problem of status collection

qclc:
After the fix, errors during status collection are shown in FederatedXXXStatus CR, e.g.:
```yaml
kind: FederatedWorkflowStatus
metadata:
  ...
apiVersion: types.kubeadmiral.io/v1alpha1
clusterStatus:
- clusterName: kubeadmiral-member-1
  collectedFields:
    metadata:
      creationTimestamp: "2023-06-13T08:36:14Z"
  error: 'Failed to get those fields: statt: not found, statusee: not found'
- clusterName: kubeadmiral-member-2
  error: 'Failed to get object from cluster, error info: failed to get object default/wf
    with client: no matches for kind "Workflow" in version "argoproj.io/v1alpha1"'
- clusterName: kubeadmiral-member-3
  error: 'Failed to get object from cluster, error info: failed to get object default/wf
    with client: no matches for kind "Workflow" in version "argoproj.io/v1alpha1"'
```


pkg/controllers/status/controller.go
```diff
@@ -554,7 +584,7 @@ func (s *StatusController) latestReplicasetDigests(
 		)
 		if err != nil {
 			keyedLogger.WithValues("cluster-name", clusterName).Error(err, "Failed to get object from cluster")
-			return nil, errors.Wrapf(err, "Failed to get %s %q from cluster %q", targetKind, key, clusterName)
+			continue
```
limhawjia, line 587: @SOF3 does it make sense to `continue` here, and possibly return an incomplete digest slice? or should we still return an error in this method and have the caller simply skip writing the latestReplicasetDigest annotation when this method errors?

the end goal is to not have an error when getting latestReplicasetDigest prevent status updates.

pkg/controllers/status/controller.go
```diff
@@ -602,7 +630,7 @@ func (s *StatusController) realUpdatedReplicas(
 		)
 		if err != nil {
 			keyedLogger.WithValues("cluster-name", clusterName).Error(err, "Failed to get object from cluster")
-			return "", errors.Wrapf(err, "Failed to get %s %q from cluster %q", targetKind, key, clusterName)
+			continue
```
limhawjia, line 633: @SOF3 same concern here, is it correct to `continue` here?

SOF3: What about just returning error here and letting the caller of `realUpdatedReplicas` log-and-ignore the error?

Actually I think we could even remove this field since we have status aggregator anyway.

pkg/controllers/status/controller.go
SOF3: Same as below, just let it error and continue executing upon latestReplicasetDigests error in the caller.

qclc: ok, this PR has been updated, please take a look.

pkg/controllers/status/controller.go
qclc: ok, this PR has been updated, please take a look.

PR #119, opened 2023-06-16
fix: stuck in cascading deletion of CR

JackZxj:
The CRD may be deleted before the CR, and then the member cluster will be stuck in the cascade deletion because the api no longer exists. So we need to ignore the NotFound error.


limhawjia (approved):
lgtm

PR #131, opened 2023-06-28
feat: informer manager

limhawjia:


pkg/util/informermanager/interface.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+)
+
+type SingleClusterInformerManager interface {
+	// Starts an informer for the given GroupResourceVersion if there isn't already one running. The new/existing
+	// informer is guaranteed to run as long as the given context remains uncancelled.
```
gary-lgy, line 29: What if `ForResource` is called multiple times with ctx of different deadlines?

limhawjia: the underlying informer will only be stopped when the last context is done

pkg/util/informermanager/interface.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+)
+
+type SingleClusterInformerManager interface {
```
SOF3, line 27: More accurate to call it "informer pool"?

PR #143, opened 2023-07-10
fix: slow reschedule on api changes and error on invalid replicaspath

limhawjia:
This PR fixes 2 issues:

1. Improves the rescheduling time during apiresource changes by reducing the default cluster status collection interval.
2. Fixes the problem where scheduler silently ignores invalid replicaspath when generating scheduling unit.

### DIFF
--- a/pkg/controllers/federatedcluster/controller.go
+++ b/pkg/controllers/federatedcluster/controller.go
@@ -108,6 +108,7 @@ func NewFederatedClusterController(
 		federatedClient:    federatedClient,
 		fedSystemNamespace: fedsystemNamespace,
 		clusterHealthCheckConfig: &ClusterHealthCheckConfig{
+			// TODO: make health check period configurable
 			Period: time.Second * 30,
 		},
 		clusterJoinTimeout: clusterJoinTimeout,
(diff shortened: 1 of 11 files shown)


pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -57,14 +57,17 @@ func schedulingUnitForFedObject(
...
-		value, err := utilunstructured.GetInt64FromPath(
+		value, exists, err := utilunstructured.GetInt64FromPath(
 			fedObject,
 			typeConfig.Spec.PathDefinition.ReplicasSpec,
 			common.TemplatePath,
 		)
 		if err != nil {
 			return nil, err
 		}
+		if !exists {
```
gary-lgy, line 68:
```suggestion
		if value == nil {
```

pkg/controllers/federatedcluster/controller.go
```diff
@@ -108,7 +108,7 @@ func NewFederatedClusterController(
 		federatedClient:    federatedClient,
 		fedSystemNamespace: fedsystemNamespace,
 		clusterHealthCheckConfig: &ClusterHealthCheckConfig{
-			Period: time.Minute,
+			Period: time.Second * 30,
```
mrlihanbo, line 111: // todo: make health check period configurable

mrlihanbo (approved):

PR #144, opened 2023-07-10
fix: use metav1.Object instead of runtime.Object for more ergonomic annotation handling

### LINKED ISSUE #99: `pkg/controllers/util/annotation/annotation.go` should use `metav1.Object` instead of `pkgruntime.Object` for more ergonomic annotation handling

kevin1689-cloud:
related to #99

### DIFF
--- a/pkg/controllers/util/annotation/annotation.go
+++ b/pkg/controllers/util/annotation/annotation.go
@@ -20,26 +20,21 @@ import (
 	"fmt"
 	"reflect"
 
-	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/runtime"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	SyncSuccessTimestamp       = "syncSuccessTimestamp"
-	LastGeneration             = "lastGeneration"
+	SyncSuccessTimestamp      = "syncSuccessTimestamp"
+	LastGeneration            = "lastGeneration"
 	LastSyncSuccessGeneration = "lastSyncSuccessGeneration"
 )
 
 // HasAnnotationKey returns true if the given object has the given annotation key in its ObjectMeta.
-func HasAnnotationKey(obj runtime.Object, key string) (bool, error) {
+func HasAnnotationKey(obj metav1.Object, key string) (bool, error) {
 	if IsNilPointer(obj) {
 		return false, fmt.Errorf("object(%T) is nil pointer", obj)
 	}
-	accessor, err := meta.Accessor(obj)
-	if err != nil {
-		return false, err
-	}
-	annotations := accessor.GetAnnotations()
+	annotations := obj.GetAnnotations()
 	if annotations == nil {
 		return false, nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/controllers/util/annotation/annotation.go
```diff
@@ -67,7 +58,7 @@ func HasAnnotationKeyValue(obj runtime.Object, key, value string) (bool, error)
 // AddAnnotation adds the given annotation key and value to the given objects ObjectMeta,
 // and overwrites the annotation value if it already exists.
 // Returns true if the object was updated.
-func AddAnnotation(obj runtime.Object, key, value string) (bool, error) {
+func AddAnnotation(obj metav1.Object, key, value string) (bool, error) {
```
gary-lgy, line 61: Besides avoiding the `meta.Accessor` overhead, another point of this refactor is to allow calling `AddAnnotation` without having to handle the returned error (given the current code paths, the old implementation of `AddAnnotation` would never return an error anyway).

The following is a possible implementation of `AddAnnotation` that satisfies the requirements:

```golang
func AddAnnotation(obj metav1.Object, key, value string) (bool) {
	has := HasAnnotationKeyValue(obj, key, value)
	if has {
		return false
	}
	annotations := obj.GetAnnotations()
	if annotations == nil {
		annotations = make(map[string]string)
	}
	annotations[key] = value
	obj.SetAnnotations(annotations)
	return true
}
```

Naturally, since the return signature has changed, all the call sites need to be changed (which is a good thing because the old error handling code is merely unnecessary bloat).

gary-lgy (changes requested):
Pls see the above comment.

kevin1689-cloud:
> Pls see the above comment.

Got it, thanks for the example which is very clear. I will working on that.

gary-lgy:
> l



> > Pls see the above comment.
> 
> Got it, thanks for the example which is very clear. I will working on that.

Meanwhile, we're in the process of a major refactor (https://github.com/kubewharf/kubeadmiral/tree/refactor/unified-federated-type) and the code on the main branch will be changed significantly by next week. I suggest delaying this fix until the refactor is done, otherwise we will have to do a lot of rebasing. :)

kevin1689-cloud:
> > > Pls see the above comment.
> > 
> > 
> > Got it, thanks for the example which is very clear. I will working on that.
> 
> Meanwhile, we're in the process of a major refactor (https://github.com/kubewharf/kubeadmiral/tree/refactor/unified-federated-type) and the code on the main branch will be changed significantly by next week. I suggest delaying this fix until the refactor is done, otherwise we will have to do a lot of rebasing. :)

Got it, how about in this pull request we only avoiding the `meta.Accessor` overhead which only need change two files:

- pkg/controllers/util/annotation/annotation.go
- pkg/controllers/util/annotation/annotation_test.go

The refactor which allow calling `AddAnnotation` without having to handle the returned error, we delaying it until the major refactor is done. After the major refactor is done, I can create another pull request to remove the returned error of `AddAnnotation` and change all the call sites.

gary-lgy:
> > > > Pls see the above comment.
> > > 
> > > 
> > > Got it, thanks for the example which is very clear. I will working on that.
> > 
> > 
> > Meanwhile, we're in the process of a major refactor ([`refactor`/unified-federated-type](https://github.com/kubewharf/kubeadmiral/tree/refactor/unified-federated-type?rgh-link-date=2023-07-11T04%3A24%3A55Z)) and the code on the main branch will be changed significantly by next week. I suggest delaying this fix until the refactor is done, otherwise we will have to do a lot of rebasing. :)
> 
> Got it, how about in this pull request we only avoiding the `meta.Accessor` overhead which only need change two files:
> 
> * pkg/controllers/util/annotation/annotation.go
> * pkg/controllers/util/annotation/annotation_test.go
> 
> For the refactor which allow calling `AddAnnotation` without having to handle the returned error, we delaying it until the major refactor is done. After the major refactor is done, I can create another pull request to remove the returned error of `AddAnnotation` and change all the call sites.
> 
> I think by this way we can make some improvement(avoiding the `meta.Accessor` overhead) and minimize the rebasing work. What do you think about that? Thanks!

Sounds good. Please fix the compile error and then LGTM.

gary-lgy (approved):

gary-lgy:
@kevin1689-cloud thanks!

PR #151, opened 2023-07-13
refactor(informer-manager): use gvk as keys and support TweakListOptions

limhawjia:


pkg/util/cluster/util.go
```diff
@@ -0,0 +1,46 @@
...
+		if condition.Type == fedcorev1a1.ClusterReady {
+			if condition.Status == corev1.ConditionTrue {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func IsClusterJoined(clusterStatus *fedcorev1a1.FederatedClusterStatus) bool {
```
gary-lgy, line 37: Are we intentionally leaving the `pkg/controllers/util` counterparts behind?

gary-lgy (approved):

pkg/util/cluster/util.go
limhawjia: Removed

PR #163, opened 2023-07-18
refactor(informer-manager): wait for cache sync before registering event handlers

limhawjia:


pkg/util/informermanager/informermanager.go
```diff
@@ -134,11 +135,15 @@ func (m *informerManager) worker(ctx context.Context) {
...
+	err, needReenqueue, delay := m.processFTC(ctx, ftc)
 	if err != nil {
 		if needReenqueue {
 			logger.Error(err, "Failed to process FederatedTypeConfig, will retry")
-			m.queue.AddRateLimited(key)
+			if delay > 0 {
+				m.queue.AddAfter(key, delay)
+			} else {
+				m.queue.AddRateLimited(key)
+			}
```
gary-lgy, line 146: Should we ignore delay if err != nil? If we don't, do want to reset backoff at line 143?

pkg/util/informermanager/informermanager.go
```diff
@@ -148,14 +153,18 @@ func (m *informerManager) worker(ctx context.Context) {
 
 	m.queue.Forget(key)
 	if needReenqueue {
-		m.queue.Add(key)
+		if delay > 0 {
+			m.queue.AddAfter(key, delay)
+		} else {
+			m.queue.AddRateLimited(key)
```
gary-lgy, line 159: `AddRateLimited` immediately after `Forget` is probably a mistake.

gary-lgy (approved):

PR #167, opened 2023-07-18
refactor: sync controller

gary-lgy:


pkg/controllers/sync/resource.go
```diff
@@ -56,94 +51,78 @@ type FederatedResource interface {
...
 	UpdateVersions(selectedClusters []string, versionMap map[string]string) error
 	DeleteVersions()
-	ComputePlacement(clusters []*fedcorev1a1.FederatedCluster) (selectedClusters sets.String, err error)
-	NamespaceNotFederated() bool
+	ComputePlacement(clusters []*fedcorev1a1.FederatedCluster) sets.Set[string]
+	SetObject(obj fedcorev1a1.GenericFederatedObject)
 }
 
+var _ FederatedResource = &federatedResource{}
+
```
limhawjia, line 61:
```suggestion
var _ version.VersionedResource = &federatedResource{}
```

pkg/controllers/sync/accessor.go
```diff
@@ -75,161 +76,127 @@ type resourceAccessor struct {
...
+	federatedObject = federatedObject.DeepCopyGenericFederatedObject()
 
-	kind := a.typeConfig.GetFederatedType().Kind
-
-	// Most federated resources have the same name as their targets.
-	targetName := common.QualifiedName{
-		Namespace: eventSource.Namespace,
-		Name:      eventSource.Name,
+	template := &unstructured.Unstructured{}
+	if err := template.UnmarshalJSON(federatedObject.GetSpec().Template.Raw); err != nil {
```
mrlihanbo, line 150: federatedObject.GetSpec().GetTemplateAsUnstructured() instead?

pkg/controllers/sync/controller.go
```diff
@@ -110,78 +120,31 @@ type SyncController struct {
...
 
 	logger klog.Logger
 }
 
-// StartSyncController starts a new sync controller for a type config
-func StartSyncController(
+// NewSyncController returns a new sync controller for the configuration
+func NewSyncController(
+	logger klog.Logger,
 	controllerConfig *util.ControllerConfig,
```
limhawjia, line 131: Would like to deprecate `ControllerConfig` now that we have `ControllerContext`. Could we extract the specific fields required as separate parameters?

pkg/controllers/sync/controller.go
```diff
@@ -190,133 +153,132 @@ func newSyncController(
...
-				s.clusterQueue.AddAfter(struct{}{}, s.clusterAvailableDelay)
+	}); err != nil {
+		return nil, fmt.Errorf("failed to add event handler generator: %w", err)
+	}
+
+	if err := s.fedInformerManager.AddClusterEventHandlers(
+		&informermanager.ClusterEventHandler{
+			Predicate: func(oldCluster, newCluster *fedcorev1a1.FederatedCluster) bool {
+				// Enqueue cluster when it's marked for deletion to ensure cascading deletion
+				return oldCluster != nil && newCluster != nil &&
```
limhawjia, line 207: Do we need to consider the case where a cluster was marked for deletion when the controller-manager was not running? In this case `oldCluster` will be nil and `newCluster.GetDeletionTimestamp().IsZero()` will be true.

pkg/controllers/sync/controller.go
```diff
@@ -190,133 +153,132 @@ func newSyncController(
...
-	}
-
-	s.fedAccessor, err = NewFederatedResourceAccessor(
-		logger, controllerConfig, typeConfig, fedNamespaceAPIResource,
-		client, s.worker.EnqueueObject, recorder)
-	if err != nil {
-		return nil, err
+		&informermanager.ClusterEventHandler{
+			Predicate: func(oldCluster, newCluster *fedcorev1a1.FederatedCluster) bool {
+				// Reconcile all federated objects when cluster becomes available
```
limhawjia, line 216:
```suggestion
				// Reconcile all federated objects when cluster becomes ready
```

pkg/controllers/sync/controller.go
```diff
@@ -190,133 +153,132 @@ func newSyncController(
...
+				return oldCluster != nil && newCluster != nil &&
+					!clusterutil.IsClusterReady(&oldCluster.Status) && clusterutil.IsClusterReady(&newCluster.Status)
+			},
+			Callback: func(cluster *fedcorev1a1.FederatedCluster) {
+				s.clusterReadinessTransitionQueue.AddAfter(struct{}{}, s.clusterAvailableDelay)
+			},
+		},
+		&informermanager.ClusterEventHandler{
+			Predicate: func(oldCluster, newCluster *fedcorev1a1.FederatedCluster) bool {
+				// Reconcile all federated objects when cluster becomes unavailable
```
limhawjia, line 226:
```suggestion
				// Reconcile all federated objects when cluster becomes unready
```

pkg/controllers/sync/resource.go
gary-lgy: Should `FederatedResource` also implement `version.VersionedResource` ?

pkg/controllers/sync/controller.go
gary-lgy: done

pkg/controllers/override/util_test.go
```diff
@@ -21,14 +21,15 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	jsonutil "k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/tools/cache"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
limhawjia, line 28: Let's fix the import grouping here

pkg/controllers/sync/resource.go
limhawjia: Yes I think so

pkg/controllers/sync/version/manager.go
```diff
@@ -70,38 +63,52 @@ type VersionManager struct {
...
-func (m *VersionManager) Sync(stopChan <-chan struct{}) {
-	versionList, ok := m.list(stopChan)
+func (m *VersionManager) Sync(ctx context.Context) {
+	versionList, ok := m.list(ctx)
 	if !ok {
 		return
 	}
-	ok = m.load(versionList, stopChan)
+	ok = m.load(ctx, versionList)
 	if !ok {
```
limhawjia, line 112: What is the point of this check? Should we be logging an error?

pkg/controllers/sync/version/manager.go
```diff
@@ -250,39 +255,33 @@ func (m *VersionManager) list(stopChan <-chan struct{}) (runtimeclient.ObjectLis
...
-	m.logger.WithValues("federated-kind", m.federatedKind).
-		V(4).Info("Version manager for federatedKind synced")
+	m.logger.Info("Version manager synced")
 	return true
 }
 
 // versionQualifiedName derives the qualified name of a version
-// resource from the qualified name of a template or target resource.
+// resource from the qualified name of a federated object.
 func (m *VersionManager) versionQualifiedName(qualifiedName common.QualifiedName) common.QualifiedName {
```
limhawjia, line 283: Is there still a need for this method now that we have declared in the API definition that the name of the `PropagatedVersion` and `ClusterPropagatedVersion` is identical to the name of a `FederatedObject` and `ClusterFederatedObject` respectively?

pkg/controllers/sync/version/manager.go
```diff
@@ -23,44 +22,38 @@ package version
...
 type VersionedResource interface {
 	FederatedName() common.QualifiedName
-	Object() *unstructured.Unstructured
+	Object() fedcorev1a1.GenericFederatedObject
 	TemplateVersion() (string, error)
 	OverrideVersion() (string, error)
+	FederatedGVK() schema.GroupVersionKind
 }
 
 type VersionManager struct {
```
limhawjia, line 54: Let's add some comments to explain what VersionManager does.

> VersionManager is used by the Sync controller to record the last synced version of a FederatedObject along with the versions of the cluster objects that were created or updated in the process. This is used to prevent sending unnecessary update requests to member clusters in subsequent reconciliations. The information is persisted in the apiserver as PropagatedVersion and ClusterPropagated CRs.
>
> In the context of the Sync controller, we identify the "version" of a FederatedObject with the hash of its template and overrides and we identify the "version" of a cluster object to be either its Generation (if available) or its ResourceVersion.
> 
> VersionManager is required because created/updated cluster objects might not match the template directly due to default values, admission plugins or webhooks. Thus we have to store the version returned by the create/update request to avoid false-positives when determining if the cluster object has diverged from the template.

pkg/controllers/sync/version/manager.go
```diff
@@ -70,38 +63,52 @@ type VersionManager struct {
...
+func (m *VersionManager) Sync(ctx context.Context) {
+	versionList, ok := m.list(ctx)
 	if !ok {
 		return
 	}
-	ok = m.load(versionList, stopChan)
+	ok = m.load(ctx, versionList)
 	if !ok {
 		return
 	}
```
limhawjia, line 114: In the comment doc for `VersionManager.Get` below, let's also explain that an empty map is returned when the given `VersionedResource` does not match the last version recorded by the `VersionManager`.

pkg/controllers/sync/resource.go
```diff
@@ -26,28 +25,24 @@ import (
...
-	"github.com/kubewharf/kubeadmiral/pkg/controllers/util/managedlabel"
-	schemautil "github.com/kubewharf/kubeadmiral/pkg/controllers/util/schema"
-	utilunstructured "github.com/kubewharf/kubeadmiral/pkg/controllers/util/unstructured"
+	annotationutil "github.com/kubewharf/kubeadmiral/pkg/util/annotation"
+	"github.com/kubewharf/kubeadmiral/pkg/util/finalizers"
+	"github.com/kubewharf/kubeadmiral/pkg/util/managedlabel"
+	overridesutil "github.com/kubewharf/kubeadmiral/pkg/util/overrides"
 )
 
 // FederatedResource encapsulates the behavior of a logical federated
```
limhawjia, line 48:
```suggestion
// FederatedResource is a wrapper for FederatedObjects and ClusterFederatedObjects that provides necessary abstractions for the Sync controller to propagate them to selected member clusters.
```

pkg/controllers/sync/resource.go
```diff
@@ -57,94 +52,83 @@ type FederatedResource interface {
 	dispatch.FederatedResourceForDispatch
```
limhawjia, line 52: Let's add a comment doc for each method

pkg/controllers/sync/placement.go
```diff
@@ -22,94 +21,24 @@ are Copyright 2023 The KubeAdmiral Authors.
...
-	}
-
-	// If both namespace and resource placement exist, the desired
-	// list of clusters is their intersection.
-	return resourceClusters.Intersection(namespaceClusters), nil
-}
-
 // computePlacement determines the selected clusters for a federated
 // resource.
 func computePlacement(
```
limhawjia, line 31: Would it be better to move these functions to `resource.go`? They are only used by `FederatedResource`.

pkg/controllers/sync/dispatch/managed.go
```diff
@@ -84,8 +72,8 @@ type ManagedDispatcher interface {
 	Update(ctx context.Context, clusterName string, clusterObj *unstructured.Unstructured)
 	VersionMap() map[string]string
 	CollectedStatus() status.CollectedPropagationStatus
-	RecordClusterError(propStatus fedtypesv1a1.PropagationStatus, clusterName string, err error)
-	RecordStatus(clusterName string, propStatus fedtypesv1a1.PropagationStatus)
+	RecordClusterError(propStatus fedcorev1a1.PropagationStatusType, clusterName string, err error)
```
limhawjia, line 75: Can we merge `RecordClusterError` and `RecordStatus` since all `RecordClusterError` is doing now is to emit an event.

pkg/controllers/sync/dispatch/managed.go
```diff
@@ -55,24 +51,16 @@ const IndexRolloutPlans = "federation_placement_rollout"
...
+	Object() fedcorev1a1.GenericFederatedObject
 	VersionForCluster(clusterName string) (string, error)
 	ObjectForCluster(clusterName string) (*unstructured.Unstructured, error)
-	ApplyOverrides(
-		obj *unstructured.Unstructured,
-		clusterName string,
-		otherOverrides fedtypesv1a1.OverridePatches,
-	) error
+	ApplyOverrides(obj *unstructured.Unstructured, clusterName string) error
 	RecordError(errorCode string, err error)
```
limhawjia, line 62: Can we merge `RecordError` and `RecordEvent` since all `RecordError` is doing now is to emit an event?

pkg/controllers/sync/dispatch/managed.go
```diff
@@ -55,24 +51,16 @@ const IndexRolloutPlans = "federation_placement_rollout"
 // interface required for dispatching operations to managed resources.
 type FederatedResourceForDispatch interface {
 	TargetName() common.QualifiedName
-	TargetKind() string
 	TargetGVK() schema.GroupVersionKind
+	TargetGVR() schema.GroupVersionResource
```
limhawjia, line 55: Are these methods necessary? We can derive them quite easily from either the FTC or the `GenericFederatedObject`

pkg/controllers/sync/accessor.go
```diff
@@ -22,50 +21,50 @@ are Copyright 2023 The KubeAdmiral Authors.
...
+	fedcorev1a1informers "github.com/kubewharf/kubeadmiral/pkg/client/informers/externalversions/core/v1alpha1"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/common"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/sync/version"
-	"github.com/kubewharf/kubeadmiral/pkg/controllers/util"
+	"github.com/kubewharf/kubeadmiral/pkg/util/eventhandlers"
+	"github.com/kubewharf/kubeadmiral/pkg/util/fedobjectadapters"
+	"github.com/kubewharf/kubeadmiral/pkg/util/informermanager"
 )
 
 // FederatedResourceAccessor provides a way to retrieve and visit
```
limhawjia, line 44:
```suggestion
// FederatedResourceAccessor provides a way to retrieve the FederatedResource representation of a FederatedObject/ClusterFederatedObject in the apiserver.
```

pkg/controllers/sync/accessor.go
```diff
@@ -22,50 +21,50 @@ are Copyright 2023 The KubeAdmiral Authors.
...
 type FederatedResourceAccessor interface {
-	Run(stopChan <-chan struct{})
+	Run(context.Context)
 	HasSynced() bool
 	FederatedResource(
 		qualifiedName common.QualifiedName,
-	) (federatedResource FederatedResource, possibleOrphan bool, err error)
-	VisitFederatedResources(visitFunc func(obj interface{}))
+	) (federatedResource FederatedResource, err error)
+	VisitFederatedResources(visitFunc func(fedcorev1a1.GenericFederatedObject))
```
limhawjia, line 52: Can we extract this method into a separate helper function or a method of the sync controller? I feel that it doesn't really make sense here.

pkg/util/meta/meta.go
```diff
@@ -18,7 +18,7 @@ This file may have been modified by The KubeAdmiral Authors
 are Copyright 2023 The KubeAdmiral Authors.
 */
 
-package util
```
limhawjia, line 21: `DeepCopyRelevantMeta` seems to be unused. Should we remove it?

limhawjia: Along with many other functions in this package: `ObjectMetaEquivalent`, `ObjectMetaAndSpecEquivalent`, `MetaAccessor`, `GetUnstructured`, `GetPluralName`.

pkg/util/propagatedversion/propagatedversion.go
```diff
@@ -19,7 +18,7 @@ This file may have been modified by The KubeAdmiral Authors
 are Copyright 2023 The KubeAdmiral Authors.
 */
 
-package util
```
limhawjia, line 22: Would it be better to move this directory under `pkg/controllers/sync`? Since it is and will probably continue to not be used by other controllers?

pkg/controllers/sync/controller.go
```diff
@@ -35,42 +33,50 @@ import (
...
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
-	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	fedclient "github.com/kubewharf/kubeadmiral/pkg/client/clientset/versioned"
```
limhawjia, line 49: Let's fix the imports formatting

pkg/controllers/sync/controller.go
```diff
@@ -82,242 +88,241 @@ const (
...
-		client,
-		configCopy,
-		&targetAPIResource,
-		func(obj pkgruntime.Object) {
-			qualifiedName := common.NewQualifiedName(obj)
-			s.worker.EnqueueWithDelay(qualifiedName, s.memberObjectEnqueueDelay)
+	if err := s.fedInformerManager.AddClusterEventHandlers(
+		&informermanager.ClusterEventHandler{
+			Predicate: func(oldCluster, newCluster *fedcorev1a1.FederatedCluster) bool {
+				// Enqueue cluster when it's added or marked for deletion to ensure cascading deletion
```
limhawjia, line 223: Will this cause us to mistakenly enqueue unjoined clusters?

limhawjia: Does this also overlap with the predicate below?

> // Trigger cascading deletion when cluster is marked for deletion

pkg/controllers/sync/controller.go
```diff
@@ -326,55 +331,69 @@ func (s *SyncController) HasSynced() bool {
 		return false
```
limhawjia, line 331: Why do we log for `ftcManager` and `fedInformerManager` but not `fedAccessor`?

pkg/controllers/sync/controller.go
```diff
@@ -926,33 +769,40 @@ func (s *SyncController) ensureRemovedOrUnmanaged(ctx context.Context, fedResour
...
-	clusters, err := s.informer.GetJoinedClusters()
+	keyedLogger := klog.FromContext(ctx)
+
+	clusters, err := s.fedInformerManager.GetJoinedClusters()
 	if err != nil {
-		return false, errors.Wrap(err, "failed to get a list of clusters")
+		return false, fmt.Errorf("failed to get a list of clusters: %w", err)
+	}
+
+	ftc, exists := s.ftcManager.GetResourceFTC(targetGVK)
```
limhawjia, line 783: Let's avoid retrieving FTC from `ftcManager` again, and use the one from the `FederatedResource`. We can pass it as an argument to this function in `deleteFromClusters`.

pkg/controllers/sync/controller.go
```diff
@@ -687,95 +629,28 @@ func (s *SyncController) setFederatedStatus(ctx context.Context, fedResource Fed
...
-	})
-	if err != nil {
-		keyedLogger.Error(err, "Failed to update syncing annotation")
-		return worker.StatusError
-	}
-
 	return worker.StatusAllOK
 }
 
 func (s *SyncController) ensureDeletion(ctx context.Context, fedResource FederatedResource) worker.Result {
```
limhawjia, line 635:
```suggestion
func (s *SyncController) handleTerminatingFederatedResource(ctx context.Context, fedResource FederatedResource) worker.Result {
```

limhawjia: Let's rename the methods used for handling deletion to be more aligned with what they do:

`ensureDeletion` -> `handleTerminatingFederatedResource`
`deleteFromClusters` -> `ensureRemovalFromClusters`
`handleDeletionFromClusters` -> `removeFromAllClusters`
`deleteFromCluster` -> `removeFromCluster`
`ensureRemovedOrUnmanaged` -> `checkObjectRemovedFromAllClusters`

pkg/controllers/sync/controller.go
```diff
@@ -496,16 +494,16 @@ func (s *SyncController) syncToClusters(ctx context.Context, fedResource Federat
...
 				// Resource is marked for deletion
-				dispatcher.RecordStatus(clusterName, fedtypesv1a1.WaitingForRemoval)
+				dispatcher.RecordStatus(clusterName, fedcorev1a1.WaitingForRemoval)
 				continue
 			}
-			if cluster.GetDeletionTimestamp() != nil && !util.IsCascadingDeleteEnabled(cluster) {
+			if cluster.GetDeletionTimestamp() != nil && !cascadingdeletion.IsCascadingDeleteEnabled(cluster) {
 				// If cluster is terminating and cascading-delete is disabled,
 				// disallow deletion to preserve cluster object.
 				// This could happen right after a cluster is deleted:
```
limhawjia, line 503:
```suggestion
				// Deletion could still happen right after a cluster is deleted:
```
Is this comment wrong? Or am I misunderstanding?

pkg/controllers/sync/controller.go
```diff
@@ -496,16 +494,16 @@ func (s *SyncController) syncToClusters(ctx context.Context, fedResource Federat
...
 				// the scheduler observes the cluster deletion and removes
 				// the placement, while the sync controller's informer is
-				// lagging behind and sees a terminating cluster.
+				// lagging behind and observes a terminating cluster.
```
limhawjia, line 506:
```suggestion
				// lagging behind and observes a non-terminating cluster.
```

pkg/controllers/sync/version/manager.go
```diff
@@ -70,38 +63,52 @@ type VersionManager struct {
 
 	versions map[string]runtimeclient.Object
```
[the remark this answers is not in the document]
gary-lgy: Would like to avoid changes to the PropagatedVersion mess as much as possible since we're uncertain what to do with it. If we decide to keep it, we could refactor it in the future for consistency.

pkg/controllers/sync/version/manager.go
gary-lgy: Looks like a redundant check. Removed it and made sure `list` and `load` logs the error correspondingly.

pkg/controllers/sync/version/manager.go
gary-lgy: Technically no. Changing the implementation of this method was the least invasive way to change PropagatedVersion naming. I can remove it if it's an issue.

pkg/controllers/sync/resource.go
```diff
@@ -306,19 +236,13 @@ func dropPodFields(obj *unstructured.Unstructured) error {
 func (r *federatedResource) ApplyOverrides(
```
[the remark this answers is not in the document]
gary-lgy: We can, but differentiating `ComputeResourceFailed` and `ApplyOverridesFailed` would be difficult, and it's outside the scope of this PR which is already huge.

pkg/controllers/sync/placement.go
gary-lgy: What's the issue with them being in a separate file? I think `resource.go` is pretty cluttered.

pkg/controllers/sync/dispatch/managed.go
gary-lgy: `RecordError` records an event with event type set as `EventTypeWarning`.

pkg/controllers/sync/dispatch/managed.go
gary-lgy: Not strictly, but I don't see a point in removing them either.

pkg/controllers/sync/accessor.go
gary-lgy: The comment for `FederatedResourceAccessor` states it "provides a way to retrieve and visit" `FederatedResource`s. It's basically a wrapper for the FO and CFO informers. The sync controller doesn't have the informer reference to implement this method.

pkg/util/meta/meta.go
gary-lgy: Removed

pkg/controllers/sync/controller.go
gary-lgy: The comment above explains it.

pkg/controllers/sync/controller.go
gary-lgy: We will enqueue unjoined clusters, but `reconcileClusterForCascadingDelete` will handle them correctly.

pkg/controllers/sync/controller.go
gary-lgy: Renamed all but `handleDeletionFromClusters` -> `removeFromAllClusters`. `handleDeletionFromClusters` invokes a custom `deletionFunc` and should be named differently from `removeFromCluster`.

pkg/controllers/sync/controller.go
gary-lgy: This block prevents the said deletion from happening. `This could happen` means `This check will succeed under the following scenario`.

pkg/controllers/sync/placement.go
limhawjia: This is a NIT, I just thought they were 2 relatively small and simple functions that look out-of-place in their own file.

pkg/controllers/sync/controller.go
limhawjia: Got it

limhawjia (approved):

PR #172, opened 2023-07-21
feat: informer manager ftc update handlers

limhawjia:
This PR adds FTC update handlers that are called every time the `InformerManager` processes an FTC. The purpose of this is to allow controllers to hook into the `InformerManager`'s view of a FTC's lifecycle. For example, reenqueue all FOs and CFOs for a FTC when it is created or when its configuration changes.

In addition, `InformerManager.HasSynced` is also modified to only return `true` when an initial snapshot of FTCs is processed. This allows controllers to avoid false "lister not found" or "ftc not found" on controller-manager startup where it processes FOs or CFOs before their corresponding FTCs are processed by `InformerManager`.

### DIFF
--- a/pkg/util/informermanager/informermanager.go
+++ b/pkg/util/informermanager/informermanager.go
@@ -281,11 +281,13 @@ func (m *informerManager) processFTCDeletionUnlocked(ctx context.Context, ftcNam
 	delete(m.eventHandlerRegistrations, ftcName)
 
 	lastObservedFTC := m.lastObservedFTCs[ftcName]
-	for _, handler := range m.ftcUpdateHandlers {
-		handler(lastObservedFTC, nil)
+	if lastObservedFTC != nil {
+		for _, handler := range m.ftcUpdateHandlers {
+			handler(lastObservedFTC, nil)
+		}
 	}
-
 	delete(m.lastObservedFTCs, ftcName)
+
 	m.initialFTCs.Delete(ftcName)
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/util/informermanager/informermanager.go
```diff
@@ -200,13 +206,18 @@ func (m *informerManager) processFTC(
...
 
-		m.lastObservedFTCs[ftcName] = ftc
 		m.informers[ftcName] = informer
 		m.informerCancelFuncs[ftcName] = cancel
 		m.eventHandlerRegistrations[ftcName] = map[*EventHandlerGenerator]cache.ResourceEventHandlerRegistration{}
 		m.lastAppliedFTCsCache[ftcName] = map[*EventHandlerGenerator]*fedcorev1a1.FederatedTypeConfig{}
 	}
 
+	for _, handler := range m.ftcUpdateHandlers {
+		handler(m.lastObservedFTCs[ftcName], ftc)
```
JackZxj, line 216: Do we need asynchronous processing here?

pkg/util/informermanager/informermanager.go
```diff
@@ -265,11 +276,18 @@ func (m *informerManager) processFTCDeletionUnlocked(ctx context.Context, ftcNam
...
 	m.gvkMapping.DeleteT1(ftcName)
 
-	delete(m.lastObservedFTCs, ftcName)
 	delete(m.informers, ftcName)
 	delete(m.informerCancelFuncs, ftcName)
 	delete(m.eventHandlerRegistrations, ftcName)
 
+	lastObservedFTC := m.lastObservedFTCs[ftcName]
+	for _, handler := range m.ftcUpdateHandlers {
+		handler(lastObservedFTC, nil)
```
gary-lgy, line 285: Should we only call `handler` when `lastObservedFTC` is not nil?

gary-lgy (approved):

pkg/util/informermanager/informermanager.go
limhawjia: I don't think it is necessary for now.

PR #179, opened 2023-07-24
refactor: status controller

mrlihanbo:


pkg/util/informermanager/interface.go
```diff
@@ -44,6 +44,17 @@ type EventHandlerGenerator struct {
 // When a FTC deletion is observed, latest will be nil.
 type FTCUpdateHandler func(lastObserved, latest *fedcorev1a1.FederatedTypeConfig)
 
+type FederatedTypeConfigManager interface {
+	// Returns the known FTC mapping for the given GVK if it exists.
+	GetResourceFTC(gvk schema.GroupVersionKind) (ftc *fedcorev1a1.FederatedTypeConfig, exists bool)
+
+	// Returns the FederatedTypeConfig lister used by the manager.
+	GetFederatedTypeConfigLister() fedcorev1a1listers.FederatedTypeConfigLister
```
limhawjia, line 52: We should probably move `AddFTCUpdateHandler` here as well.

limhawjia: I think if `GetFederatedTypeConfigLister` is unused, we can remove this method from both `FederatedTypeConfigManager` and `InformerManager`.

The underlying FTC lister might not match the `InformerManager`'s view of FTCs as the `InformerManager`'s view is still subject to its own internal processing and tends to lag behind.

cc @gary-lgy

cmd/controller-manager/app/options/options.go
```diff
@@ -50,9 +50,10 @@ type Options struct {
 	LogVerbosity    int
 	KlogVerbosity   int
 
-	NSAutoPropExcludeRegexp string
-	CreateCRDsForFTCs       bool
-	ClusterJoinTimeout      time.Duration
+	NSAutoPropExcludeRegexp  string
+	CreateCRDsForFTCs        bool
```
limhawjia, line 54: This option is not needed anymore.

pkg/controllers/status/controller.go
```diff
@@ -68,181 +73,148 @@ const (
...
+		reconcileOnClusterChangeDelay:  time.Second * 3,
+		memberObjectEnqueueDelay:       memberObjectEnqueueDelay,
+		metrics:                        metrics,
+		logger:                         logger.WithValues("controller", StatusControllerName),
+		eventRecorder:                  eventsink.NewDefederatingRecorderMux(kubeClient, StatusControllerName, 4),
 	}
 
 	s.worker = worker.NewReconcileWorker(
+		StatusControllerName,
+		nil,
```
limhawjia, line 146:
```suggestion
		common.NewQualifiedName,
```

pkg/controllers/status/controller.go
```diff
@@ -68,181 +73,148 @@ const (
...
+		workerCount,
+		metrics,
 	)
 
 	// Build queue for triggering cluster reconciliations.
 	s.clusterQueue = workqueue.NewNamedDelayingQueue("status-controller-cluster-queue")
 
-	// Start informers on the resources for the federated type
-	enqueueObj := s.worker.EnqueueObject
+	handler := util.NewTriggerOnAllChanges(func(o pkgruntime.Object) {
```
limhawjia, line 156:
```suggestion
	handler := util.NewTriggerOnAllChanges(s.worker.EnqueueObject)
```

pkg/controllers/status/controller.go
```diff
@@ -68,181 +73,148 @@ const (
...
-		configCopy,
-		&targetAPIResource,
-		func(obj pkgruntime.Object) {
-			qualifiedName := common.NewQualifiedName(obj)
-			s.worker.EnqueueWithDelay(qualifiedName, s.memberObjectEnqueueDelay)
+	s.collectedStatusInformer.Informer().AddEventHandler(handler)
+	s.clusterCollectedStatusInformer.Informer().AddEventHandler(handler)
+
+	if err := s.fedInformerManager.AddEventHandlerGenerator(&informermanager.EventHandlerGenerator{
+		Predicate: informermanager.RegisterOncePredicate,
```
limhawjia, line 167: Since status collection can be enabled/disabled in the FTC, I think we need to check for this in the predicate.

pkg/controllers/status/controller.go
```diff
@@ -68,181 +73,148 @@ const (
...
+		Generator: func(ftc *fedcorev1a1.FederatedTypeConfig) cache.ResourceEventHandler {
+			return eventhandlers.NewTriggerOnAllChanges(func(o pkgruntime.Object) {
```
limhawjia, line 169: Based on the same reasoning as above, we should return `nil` here when status collection is disabled.

pkg/controllers/status/controller.go
```diff
@@ -253,53 +225,67 @@ func (s *StatusController) Run(stopChan <-chan struct{}) {
...
 // Check whether all data stores are in sync. False is returned if any of the informer/stores is not yet
 // synced with the corresponding api server.
 func (s *StatusController) HasSynced() bool {
-	if !s.informer.ClustersSynced() {
-		s.logger.V(3).Info("Cluster list not synced")
-		return false
-	}
-	if !s.federatedController.HasSynced() {
-		s.logger.V(3).Info("Federated type not synced")
+	if !s.ftcManager.HasSynced() {
```
limhawjia, line 245: Why is this log only necessary for `ftcManager` and `fedInformerManager`?

pkg/controllers/status/controller.go
```diff
@@ -253,53 +225,67 @@ func (s *StatusController) Run(stopChan <-chan struct{}) {
 		}
 	}()
 
-	if !cache.WaitForNamedCacheSync(s.name, stopChan, s.HasSynced) {
+	if !cache.WaitForNamedCacheSync(StatusControllerName, ctx.Done(), s.HasSynced) {
+		s.logger.Error(nil, "Timed out waiting for cache sync")
 		return
 	}
```
limhawjia, line 232: Let's add a log here as well - `s.logger.Info("Caches are synced")`

pkg/controllers/status/controller.go
```diff
@@ -310,39 +296,69 @@ func (s *StatusController) reconcile(qualifiedName common.QualifiedName) (reconc
...
+	template, err := fedObject.GetSpec().GetTemplateAsUnstructured()
+	if err != nil {
+		keyedLogger.Error(err, "Failed to unmarshal template")
+		return worker.StatusError
+	}
+
+	templateGVK := template.GroupVersionKind()
+	targetIsDeployment := templateGVK == appsv1.SchemeGroupVersion.WithKind(common.DeploymentKind)
+
+	templateQualifiedName := common.QualifiedName{Namespace: template.GetNamespace(), Name: template.GetName()}
```
limhawjia, line 331:
```suggestion
	templateQualifiedName := common.NewQualifiedName(template)
```

pkg/controllers/status/controller.go
```diff
@@ -310,39 +296,69 @@ func (s *StatusController) reconcile(qualifiedName common.QualifiedName) (reconc
...
+	}
+
 	clusterNames, err := s.clusterNames()
 	if err != nil {
 		keyedLogger.Error(err, "Failed to get cluster list")
 		return worker.Result{RequeueAfter: &s.clusterAvailableDelay}
 	}
 
-	clusterStatus := s.clusterStatuses(ctx, fedObject, clusterNames, qualifiedName)
+	clusterStatus := s.clusterStatuses(ctx, fedObject, templateQualifiedName, templateGVK, typeConfig, clusterNames)
```
limhawjia, line 344:
```suggestion
	clusterStatuses := s.clusterStatuses(ctx, fedObject, templateQualifiedName, templateGVK, typeConfig, clusterNames)
```

pkg/controllers/status/controller.go
```diff
@@ -495,24 +457,27 @@ func (s *StatusController) clusterNames() ([]string, error) {
...
 	// collect errors during status collection and record them as event
-	errList := []string{}
+	var errList []string
 
 	for _, clusterName := range clusterNames {
-		resourceClusterStatus := util.ResourceClusterStatus{ClusterName: clusterName}
+		resourceClusterStatus := fedcorev1a1.CollectedFieldsWithCluster{Cluster: clusterName}
 
-		clusterObj, exist, err := util.GetClusterObject(
+		clusterObj, exist, err := informermanager.GetClusterObject(
```
limhawjia, line 474: How about reenqueuing if the informer is not yet synced instead? If we create a new FTC,  the informer for the source type won't be synced for a while + all objects of the source type will be enqueued at the same time. This would lead to a burst of requests to apiserver, potentially affecting the apiserver or causing ourselves to be rate-llimited.
cc @gary-lgy

pkg/controllers/status/controller.go
limhawjia: Realized it doesn't affect correctness since we check for status collection enablement during reconcile anyway. But it would be good to reduce unnecessary processing as well.

pkg/util/informermanager/interface.go
limhawjia: Will handle this in a subsequent pr.

pkg/controllers/status/controller.go
limhawjia: We can explore this issue in a new PR instead. This will do for now.

pkg/controllers/status/controller.go
```diff
@@ -68,181 +73,168 @@ const (
...
+
+	if err := s.fedInformerManager.AddEventHandlerGenerator(&informermanager.EventHandlerGenerator{
+		Predicate: func(lastApplied, latest *fedcorev1a1.FederatedTypeConfig) bool {
+			return lastApplied.IsStatusCollectionEnabled() != latest.IsStatusCollectionEnabled()
 		},
-		&util.ClusterLifecycleHandlerFuncs{
-			ClusterAvailable: func(cluster *fedcorev1a1.FederatedCluster) {
-				// When new cluster becomes available process all the target resources again.
+		Generator: func(ftc *fedcorev1a1.FederatedTypeConfig) cache.ResourceEventHandler {
+			if ftc.Spec.StatusCollection == nil || !ftc.Spec.StatusCollection.Enabled {
```
limhawjia, line 185:
```suggestion
			if !ftc.IsStatusCollectionEnabled() {
```

limhawjia (approved):

PR #191, opened 2023-08-07
fix: bugfix for status, statusaggregator and auto-migration

JackZxj:
1. remove revisionHistory of FTC
2. fix failure to get podList in auto-migration
3. fix bug that FTC updating does not trigger status controller
4. ignore resources with NoFederatedResource annotation in statusaggregator

### DIFF
--- a/pkg/controllers/status/controller.go
+++ b/pkg/controllers/status/controller.go
@@ -182,7 +182,9 @@ func NewStatusController(
 			if lastApplied == nil || latest == nil {
 				return true
 			}
-			return lastApplied.IsStatusCollectionEnabled() != latest.IsStatusCollectionEnabled()
+			return lastApplied.IsStatusCollectionEnabled() != latest.IsStatusCollectionEnabled() ||
+				(lastApplied.Spec.StatusCollection != nil && latest.Spec.StatusCollection != nil &&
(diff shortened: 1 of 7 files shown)


mrlihanbo (approved):

pkg/controllers/status/controller.go
```diff
@@ -233,6 +233,17 @@ func NewStatusController(
 		return nil, fmt.Errorf("failed to add cluster event handler: %w", err)
 	}
 
+	if err := s.ftcManager.AddFTCUpdateHandler(func(lastObserved, latest *fedcorev1a1.FederatedTypeConfig) {
+		if latest == nil || !latest.IsStatusCollectionEnabled() {
```
limhawjia, line 237: I think this overlaps with the `Predicate` passed to `AddEventHandlerGenerators` above. `AddEventHandlerGenerator` will also trigger a reenqueue of all objects as the newly registered eventhandler will receive an add event for all existing objects in the store.

limhawjia: We probably only need to keep the `!reflect.DeepEqual(lastObserved.Spec.StatusCollection.Fields, latest.Spec.StatusCollection.Fields)` check here. Alternatively, we can move this check to the `Predicate`s above.

mrlihanbo (approved):

limhawjia (approved):

PR #194, opened 2023-08-07
chore:add log when federated cluster is not ready

wy-lucky:
Record the cluster health check body when federated cluster is not ready, so we can analyze the cause of the anomaly.

### DIFF
--- a/pkg/controllers/federatedcluster/clusterstatus.go
+++ b/pkg/controllers/federatedcluster/clusterstatus.go
@@ -80,6 +80,7 @@ func (c *FederatedClusterController) collectIndividualClusterStatus(
 
 	discoveryClient := clusterKubeClient.Discovery()
 
+	oldClusterStatus := cluster.Status.DeepCopy()
 	cluster = cluster.DeepCopy()
 	conditionTime := metav1.Now()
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -163,6 +163,7 @@ func checkReadyByHealthz(
 	if strings.EqualFold(string(body), "ok") {
 		clusterReadyStatus = corev1.ConditionTrue
 	} else {
+		logger.V(3).WithValues("body", body).Info("Cluster is not ready")
```
gary-lgy, line 166: Should we also log if the request fails?

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -144,6 +144,10 @@ func (c *FederatedClusterController) collectIndividualClusterStatus(
 		return 0, fmt.Errorf("failed to update cluster status: %w", err)
 	}
 
+	if readyStatus != corev1.ConditionTrue {
+		c.eventRecorder.Eventf(cluster, ClusterHealthzNotOKReason, ClusterHealthzNotOKMessage, "Cluster is not ready")
```
gary-lgy, line 149: We should only log when there's a state transition. Otherwise an unready cluster will keep emitting events.

wy-lucky: done

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -144,9 +145,34 @@ func (c *FederatedClusterController) collectIndividualClusterStatus(
...
 	}
 
+	if isReadyStatusChanged(oldClusterStatus, readyStatus) {
+		switch readyStatus {
+		case corev1.ConditionTrue:
+			c.eventRecorder.Eventf(cluster, readyReason, readyMessage, "Cluster is ready")
+		case corev1.ConditionFalse:
+			c.eventRecorder.Eventf(cluster, readyReason, readyMessage, "Cluster is not ready")
+		case corev1.ConditionUnknown:
+			c.eventRecorder.Eventf(cluster, readyReason, readyMessage, "Cluster is not reachable")
```
gary-lgy, line 155:
```suggestion
			c.eventRecorder.Eventf(cluster, readyReason, readyMessage, "Cluster ready state is unknown")
```

gary-lgy (approved):

PR #197, opened 2023-08-09
Support gpu scheduling

Poor12:
Pass the gpu resourceQuest into schedulingUnit so that we can filter and score the cluster based on gpu resource quest of workloads.
At present, the interpret of the PodTemplate path of K8s native resources is built in, and it can be put into FTC for subsequent consideration.

Refer to https://github.com/kubewharf/kubeadmiral/blob/7ede6f4b715d2b3861ce8a6b4f681cc229d86cfe/pkg/util/pod/pod.go#L38

### DIFF
--- a/pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
@@ -49,9 +49,14 @@ func (pl *ClusterResourcesLeastAllocated) Score(
 		return 0, framework.NewResult(framework.Error, err.Error())
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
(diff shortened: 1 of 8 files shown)


Poor12:
Test Result:
When apply a deploy whose resource quest is more than available resources in clusters
```
I0809 04:24:28.620856       1 generic_scheduler.go:164] "Cluster doesn't fit" controller="scheduler" key="default/echo-server-deployments.apps" source-gvk="apps/v1, Kind=Deployment" ftc="deployments.apps" policy="default/policy-all-clusters" name="kubeadmiral-member-1" reason="Insufficient cpu"
I0809 04:24:28.620879       1 generic_scheduler.go:164] "Cluster doesn't fit" controller="scheduler" key="default/echo-server-deployments.apps" source-gvk="apps/v1, Kind=Deployment" ftc="deployments.apps" policy="default/policy-all-clusters" name="kubeadmiral-member-2" reason="Insufficient cpu"
I0809 04:24:28.620893       1 generic_scheduler.go:164] "Cluster doesn't fit" controller="scheduler" key="default/echo-server-deployments.apps" source-gvk="apps/v1, Kind=Deployment" ftc="deployments.apps" policy="default/policy-all-clusters" name="kubeadmiral-member-3" reason="Insufficient cpu"
```

PR #213, opened 2023-08-22
fix: e2e tests

limhawjia:
This PR fixes the existing e2e tests and sets up a ci workflow to run e2e tests.


.github/workflows/ci.yml
```diff
@@ -28,6 +28,30 @@ jobs:
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: "1.20"
+
+      - name: Install Go dependencies
+        run: |
+          go install github.com/onsi/ginkgo/v2/ginkgo
+          go install sigs.k8s.io/kwok/cmd/{kwok,kwokctl}@v0.3.0
```
SOF3, line 44: kwok and kind both required for the same e2e test? but you only use kind cluster provider below?

SOF3:
nit: adding tests to CI does not impact features; commit title prefix should be `chore`/`ci` instead?

PR #224, opened 2023-08-29
fix: adopt exist nodeport service

mrlihanbo:


pkg/controllers/sync/dispatch/managed.go
```diff
@@ -220,6 +220,7 @@ func (d *managedDispatcherImpl) Create(ctx context.Context, clusterName string)
 		// TODO Figure out why attempting to create a namespace that
 		// already exists indicates ServerTimeout instead of AlreadyExists.
 		alreadyExists := apierrors.IsAlreadyExists(err) ||
+			d.fedResource.TargetGVK() == corev1.SchemeGroupVersion.WithKind(common.ServiceKind) && apierrors.IsInvalid(err) ||
```
JackZxj, line 223: Could you leave some comments explaining why the service needs to be handled this way?

JackZxj (approved):

pkg/controllers/sync/dispatch/managed.go
```diff
@@ -220,6 +220,9 @@ func (d *managedDispatcherImpl) Create(ctx context.Context, clusterName string)
 		// TODO Figure out why attempting to create a namespace that
```
limhawjia, line 220: The logic is fine, but I think we can try to rewrite the code here in a clearer way, perhaps something like:

```
		// If the object exists, we may attempt to adopt the resource. We check if the object exists by examining the
		// returned error.
		var needsExistenceRecheck bool
		switch {
		case apierrors.IsAlreadyExists(createErr):
			needsExistenceRecheck = false
		case d.fedResource.TargetGVK() == corev1.SchemeGroupVersion.WithKind(common.NamespaceKind) && apierrors.IsServerTimeout(createErr):
			// For namespaces, ServerTimeout is returned if object exists. We differentiate with the subsequent GET.
			needsExistenceRecheck = true
		case d.fedResource.TargetGVK() == corev1.SchemeGroupVersion.WithKind(common.ServiceKind) && apierrors.IsInvalid(createErr):
			// For services, Invalid is returned if the object exists. We differentiate with the subsequent GET.
			needsExistenceRecheck = true
		default:
			return d.recordOperationError(ctxWithTimeout, fedtypesv1a1.CreationFailed, clusterName, op, createErr)
		}

		getErr := client.Get(ctxWithTimeout, obj, obj.GetNamespace(), obj.GetName())
		if needsExistenceRecheck && apierrors.IsNotFound(getErr) {
			// The previous create error should not be an AlreadyExists error, skip trying to adopt.
			//
			// Note: there is a chance that the previous error is still an AlreadyExists error and the object was
			// simply deleted in between the 2 requests.
			return d.recordOperationError(ctxWithTimeout, fedtypesv1a1.CreationFailed, clusterName, op, createErr)
		} else if getErr != nil {
			wrappedErr := errors.Wrapf(getErr, "failed to retrieve object potentially requiring adoption")
			return d.recordOperationError(ctxWithTimeout, fedtypesv1a1.RetrievalFailed, clusterName, op, wrappedErr)
		}
```

mrlihanbo: done

limhawjia (approved):

PR #225, opened 2023-08-29
feat(sync): optimize the cascading deletion of managed member cluster resources

qclc:


qclc:
> The optimization logic seems fine but I have one comment regarding the implementation:
> 
> We probably do not need to introduce a `FederatedObjectInformer` or `ClusterFederatedObjectInformer` (correct me if I'm wrong). We can simply loop through the managed cluster objects and get its corresponding federated object through the `FederatedResourceAccessor`. See the method `FederatedObject.Object()`. This should simplify the code and avoid introducing more dependencies to the sync controller.
 
I think it's a good idea. The code has already been updated.

pkg/controllers/sync/controller.go
```diff
@@ -550,7 +550,10 @@ func (s *SyncController) syncToClusters(
 		isSelectedCluster := selectedClusterNames.Has(clusterName)
 		isCascadingDeletionTriggered := cluster.GetDeletionTimestamp() != nil &&
 			cascadingdeletion.IsCascadingDeleteEnabled(cluster)
-		shouldBeDeleted := !isSelectedCluster || isCascadingDeletionTriggered
+		if isCascadingDeletionTriggered {
+			continue
```
limhawjia, line 554: i'm not sure if it's sound to immediately skip processing a cluster at this point. here are some questions to consider:

1. if a cluster is not selected but previously contained the object, we now defer its deletion to the cascading-deletion process while the current sync reports a success. is this acceptable to the user? (sync success is reported even though it does not match the desired placement)
2. if a cluster is not selected but previously contained the object (same scenario as 1), what is the new FederatedObject status after the sync is complete? should it (and does it) reflect the status of the skipped cluster in any way? 
3. after an object is deleted by the cascading-delete loop, is the corresponding FederatedObject's status updated to reflect the new state?

limhawjia: cc @gary-lgy @mrlihanbo

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1104,107 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
-			client, exists := s.fedInformerManager.GetClusterDynamicClient(cluster.Name)
-			if !exists {
-				remainingByGVK[gvk] = "Unknown (cluster client does not exist)"
-				continue
+
+			// update some cluster objects before deleting them
+			var updateFailedObjs []string
+			for i := range clusterObjs {
+				clusterObj := clusterObjs[i].DeepCopy()
+				keyedLogger := klog.FromContext(ctx).
```
JackZxj, line 1146: This keyedLogger should be generated from logger rather then ctx, and the gvk infor should be included

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1225,138 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+		// the CRD may be deleted before the CR, and then the member cluster will be stuck
+		// in the cascade deletion because the api no longer exists. So we need to ignore
+		// the NoMatch and NotFound error.
+		// Whether to return NoMatch or NotFound depends on whether the client has visited CR,
+		// if so, returns NotFound (because the client has a scheme cache), otherwise returns NoMatch.
+		if err != nil && !meta.IsNoMatchError(err) && !apierrors.IsNotFound(err) {
+			return nil, nil, err
+		}
+
+		clusterObjs = make([]*unstructured.Unstructured, len(objects.Items))
```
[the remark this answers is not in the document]
qclc: The code has been updated.

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1225,138 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+	var clusterObjs []*unstructured.Unstructured
+	var namespaces []string
+	resourceLister, hasSynced, exists := s.fedInformerManager.GetResourceLister(ftc.GetSourceTypeGVK(), clusterName)
+	if !exists {
+		return nil, nil, fmt.Errorf("failed to get dynamic client")
+	}
+
+	// If cluster cache is synced, we check the store.
+	// Otherwise, we will have to issue a list request.
+	labelSelector := labels.SelectorFromSet(labels.Set{managedlabel.ManagedByKubeAdmiralLabelKey: managedlabel.ManagedByKubeAdmiralLabelValue})
```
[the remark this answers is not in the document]
qclc: The code has been updated.

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1225,138 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
 
 	return worker.StatusAllOK
 }
+
+func (s *SyncController) listManagedClusterObjects(
+	ctx context.Context,
+	ftc *fedcorev1a1.FederatedTypeConfig,
+	clusterName string,
+) ([]*unstructured.Unstructured, []string, error) {
```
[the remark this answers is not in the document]
qclc: The code has been updated.

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1104,107 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+	for i := range ftcs {
+		wg.Add(1)
+		go func(ftc *fedcorev1a1.FederatedTypeConfig) {
+			defer wg.Done()
+			gvk := ftc.GetSourceTypeGVK().String()
+
+			// get all managed cluster objects according to ftc type
+			clusterObjs, namespaces, err := s.listManagedClusterObjects(ctx, ftc, cluster.Name)
+			if err != nil {
+				logger.Error(err, "Failed to list clusterObjects")
```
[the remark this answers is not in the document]
qclc: The code has been updated.

pkg/controllers/sync/controller.go
qclc: The code has been updated.

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1104,113 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+				remainingByGVK[gvk] = fmt.Sprintf("failed to list cluster objects for %s", gvk)
```
JackZxj, line 1132: A risk of concurrent map writes

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1231,143 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+	ctx context.Context,
+	ftc *fedcorev1a1.FederatedTypeConfig,
+	namespaces []string,
+	client dynamic.Interface,
+	clusterName string,
+) error {
+	sourceGVR := ftc.GetSourceTypeGVR()
+	deletionPropagation := metav1.DeletePropagationBackground
+	var methodNotSupportedErr error
+	if ftc.GetNamespaced() {
```
JackZxj, line 1305: Can we consider the namespaces of the cluster-scoped resources as `[]string{corev1.NamespaceAll}`, so that the two operations can be unified?

pkg/controllers/sync/controller.go
qclc: The latest code adds a FederatedObject status update operation in the `syncToClusters` function: 
- If cascading deletion is triggered, wait for `reconcileClusterForCascadingDeletion` to perform the deletion operation. 
- If the resource has not been deleted yet, update the FederatedObject status to wait for deletion.

```go
if isCascadingDeletionTriggered {
	if clusterObj != nil && clusterObj.GetDeletionTimestamp() == nil {
		dispatcher.RecordStatus(clusterName, fedcorev1a1.WaitingForRemoval)
	}
	continue
}
```

pkg/controllers/sync/controller.go
qclc: The code has been updated.

pkg/controllers/sync/controller.go
qclc: If the gvk of each ftc is different, each thread writes to different keys in the map, and there should be no conflicts.

pkg/controllers/sync/controller.go
limhawjia: This resolves the first 2 issues I have, but what about after the object is deleted in the cascading-delete loop? Is the corresponding FederatedObject's status updated to remove the cluster?

qclc: When the member cluster is completely deleted, the FederatedObject object will be re-queued for the sync process. Since the member cluster was deleted, it won't be included in this synchronization status set. 
(Since FederatedObject's status is fully updated every time, the cluster status not included in the status set will be deleted.)

limhawjia (approved):

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
 		return worker.StatusAllOK
 	}
 
-	// cascading-delete is enabled, wait for member objects to be deleted
+	if !clusterutil.IsClusterReady(&cluster.Status) && cascadingdeletion.IsCascadingDeleteEnabled(cluster) {
```
gary-lgy, line 1114: Is there a need to check for `cascadingdeletion.IsCascadingDeleteEnabled(cluster)` here?

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+		return worker.Result{RequeueAfter: &s.cascadingDeletionRecheckDelay}
```
gary-lgy, line 1115: Let's print a log here.

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1245,135 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+	ctx context.Context,
+	ftc *fedcorev1a1.FederatedTypeConfig,
+	clusterName string,
+) ([]*unstructured.Unstructured, sets.Set[string], error) {
+	clusterObjs := []*unstructured.Unstructured{}
+	namespaces := sets.New[string]()
+
+	resourceLister, hasSynced, exists := s.fedInformerManager.GetResourceLister(ftc.GetSourceTypeGVK(), clusterName)
+	if !exists {
+		return nil, nil, fmt.Errorf("failed to get dynamic client")
```
gary-lgy, line 1259: Wrong error message

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1245,135 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+	}
+
+	// If cluster cache is synced, we check the store.
+	// Otherwise, we will have to issue a list request.
+	if hasSynced() {
+		objects, err := resourceLister.List(labels.Everything())
+		if err != nil {
+			return nil, nil, err
```
gary-lgy, line 1267: Let's wrap the errors in this function for more informative messages.

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+				syncMap.Store(gvk, "failed to get cluster client")
+				return
 			}
-		}
+
+			// update some cluster objects before deleting them
+			var updateFailedObjs []string
+			for i := range clusterObjs {
+				clusterObj := clusterObjs[i].DeepCopy()
+				keyedLogger := logger.WithValues(
```
gary-lgy, line 1157: Let's use  `logging.InjectLoggerValues`

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
-		if !exists {
-			remainingByGVK[gvk] = fmt.Sprintf("failed to get resource lister for %s", gvk)
-			continue
-		}
+	// concurrent operations for each ftc
+	wg := sync.WaitGroup{}
+	syncMap := sync.Map{}
+	for i := range ftcs {
+		wg.Add(1)
+		go func(ftc *fedcorev1a1.FederatedTypeConfig) {
```
gary-lgy, line 1131: Shall we extract this into a separate function?

pkg/controllers/sync/controller.go
```diff
@@ -1186,3 +1245,135 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+				ctx,
+				clusterObj.GetName(),
+				metav1.DeleteOptions{PropagationPolicy: &deletionPropagation},
+			)
+			if err != nil && !apierrors.IsNotFound(err) {
+				failedObjs = append(failedObjs, fmt.Sprintf("%s/%s", clusterObj.GetNamespace(), clusterObj.GetName()))
+			}
+		}
+		if len(failedObjs) != 0 {
+			return fmt.Errorf("failed to delete those cluster objects: %v", failedObjs)
```
gary-lgy, line 1375:
```suggestion
			return fmt.Errorf("failed to delete some cluster objects: %v", failedObjs)
```

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+
+			// if the update operation fails, do not continue the delete operation,
+			// but return directly to avoid accidentally deleting any orphaned cluster objects.
+			if len(updateFailedObjs) > 0 {
+				syncMap.Store(gvk, fmt.Sprintf("failed to update those cluster objects for %s: %v", gvk, updateFailedObjs))
+				return
+			}
+
+			// if the update operation is successful, all managed cluster objects will be deleted.
+			if err := s.deleteClusterObjects(ctx, ftc, namespaces.UnsortedList(), client, cluster.Name); err != nil {
```
gary-lgy, line 1212: Why don't we pass `clusterObject` into this function here? Listing again would be very expensive.

pkg/controllers/sync/controller.go
```diff
@@ -1104,58 +1111,120 @@ func (s *SyncController) reconcileClusterForCascadingDeletion(
...
+				}
+				if fedObj == nil {
+					err := fmt.Errorf("the federated object %s is not found", fedObjName)
+					keyedLogger.Error(err, "Failed to get corresponding federated object")
+					updateFailedObjs = append(updateFailedObjs, namespacedName)
+					continue
+				}
+
+				// remove the finalizer added by admiral
+				needUpdate, err := dispatch.RemoveRetainObjectFinalizer(clusterObj)
```
gary-lgy, line 1182: Should this method be extracted out from the dispatch package?

pkg/controllers/sync/controller.go
qclc: The check here is unnecessary and has been removed.

pkg/controllers/sync/controller.go
qclc: A log has been added here.

pkg/controllers/sync/controller.go
qclc: The error message has been corrected.

pkg/controllers/sync/controller.go
qclc: The error message has been wrapped.

pkg/controllers/sync/controller.go
qclc: The code here has been changed to use `logging.InjectLoggerValues`.

pkg/controllers/sync/controller.go
qclc: This should be considered the main logic, but it is encapsulated in the style of an anonymous function for concurrency. I think it does not need to be extracted separately.

pkg/controllers/sync/controller.go
qclc: The code has been updated.

pkg/controllers/sync/controller.go
qclc: The code has been updated.

pkg/controllers/sync/controller.go
qclc: The method has been extracted.

gary-lgy:
Superseded by #230

PR #226, opened 2023-09-04
Support gpu scheduling

Poor12:
This PR will:
* Pass in the gpu resource request into the scheduling unit
* `ClusterResourceFit` supports filter of gpu
* `MostAllocatable` and `LeastAllocatable` supports gpu resources, and the default weight map will be `cpu: memory: gpu=1:1:4`
* `BalancedAllocatable` supports comparison of GPU resources, based on standard deviation algorithm

### DIFF
--- a/pkg/controllers/scheduler/framework/plugins/clusterresources/balanced_allocation.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterresources/balanced_allocation.go
@@ -24,8 +24,6 @@ import (
 	"context"
 	"math"
 
-	corev1 "k8s.io/api/core/v1"
-
 	fedcorev1a1 "github.com/kubewharf/kubeadmiral/pkg/apis/core/v1alpha1"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework"
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework/plugins/names"
@@ -47,32 +45,47 @@ func (pl *ClusterResourcesBalancedAllocation) Score(
 	su *framework.SchedulingUnit,
 	cluster *fedcorev1a1.FederatedCluster,
 ) (int64, *framework.Result) {
+	var totalFraction float64
+
 	err := framework.PreCheck(ctx, su, cluster)
 	if err != nil {
 		return 0, framework.NewResult(framework.Error, err.Error())
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
-		allocatable[resource], requested[resource] = calculateResourceAllocatableRequest(su, cluster, resource)
+	resources := getRelevantResources(su)
+	allocatable, requested := getAllocatableAndRequested(su, cluster, resources)
+
+	fractions := make([]float64, 0, len(requested))
+	for i := range requested {
+		fraction := fractionOfCapacity(requested[i], allocatable[i])
+		// This to find a cluster which has most balanced resource usage.
+		if fraction >= 1 {
+			// if requested >= capacity, the corresponding host should never be preferred.
+			return 0, framework.NewResult(framework.Success)
+		}
+		totalFraction += fraction
+		fractions = append(fractions, fraction)
 	}
 
-	cpuFraction := fractionOfCapacity(requested[corev1.ResourceCPU], allocatable[corev1.ResourceCPU])
-	memoryFraction := fractionOfCapacity(requested[corev1.ResourceMemory], allocatable[corev1.ResourceMemory])
-	// This to find a node which has most balanced CPU, memory and volume usage.
-	if cpuFraction >= 1 || memoryFraction >= 1 {
-		// if requested >= capacity, the corresponding host should never be preferred.
-		return 0, framework.NewResult(framework.Success)
+	std := 0.0
+
+	// For most cases, resources are limited to cpu and memory, the std could be simplified to std := (fraction1-fraction2)/2
+	// len(fractions) > 2: calculate std based on the well-known formula - root square of Σ((fraction(i)-mean)^2)/len(fractions)
+	// Otherwise, set the std to zero is enough.
+	if len(fractions) == 2 {
+		std = math.Abs((fractions[0] - fractions[1]) / 2)
+	} else if len(fractions) > 2 {
+		mean := totalFraction / float64(len(fractions))
+		var sum float64
+		for _, fraction := range fractions {
+			sum += (fraction - mean) * (fraction - mean)
+		}
+		std = math.Sqrt(sum / float64(len(fractions)))
 	}
 
-	// Upper and lower boundary of difference between cpuFraction and memoryFraction are -1 and 1
-	// respectively. Multiplying the absolute value of the difference by 10 scales the value to
-	// 0-10 with 0 representing well balanced allocation and 10 poorly balanced. Subtracting it from
-	// 10 leads to the score which also scales from 0 to 10 while 10 representing well balanced.
-	diff := math.Abs(cpuFraction - memoryFraction)
-	score := int64((1 - diff) * float64(framework.MaxClusterScore))
-	return score, framework.NewResult(framework.Success)
+	// STD (standard deviation) is always a positive value. 1-deviation lets the score to be higher for cluster which has least deviation and
+	// multiplying it with `MaxClusterScore` provides the scaling factor needed.
+	return int64((1 - std) * float64(framework.MaxClusterScore)), framework.NewResult(framework.Success)
 }
 
 // ScoreExtensions of the Score plugin.
--- a/pkg/controllers/scheduler/framework/plugins/clusterresources/fit.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterresources/fit.go
@@ -173,3 +173,29 @@ func calculateResourceAllocatableRequest(
 
 	return 0, 0
 }
+
+func getRelevantResources(su *framework.SchedulingUnit) []corev1.ResourceName {
+	resources := make([]corev1.ResourceName, 0, len(framework.DefaultRequestedRatioResources))
+	for resourceName := range framework.DefaultRequestedRatioResources {
+		if resourceName == corev1.ResourceCPU || resourceName == corev1.ResourceMemory ||
+			su.ResourceRequest.HasScalarResource(resourceName) {
+			resources = append(resources, resourceName)
+		}
+	}
+
+	return resources
+}
+
+func getAllocatableAndRequested(
+	su *framework.SchedulingUnit,
+	cluster *fedcorev1a1.FederatedCluster,
+	resources []corev1.ResourceName,
+) (framework.ResourceToValueMap, framework.ResourceToValueMap) {
+	requested := make(framework.ResourceToValueMap, len(resources))
+	allocatable := make(framework.ResourceToValueMap, len(resources))
+	for _, resourceName := range resources {
+		allocatable[resourceName], requested[resourceName] = calculateResourceAllocatableRequest(su, cluster, resourceName)
+	}
+
+	return allocatable, requested
+}
--- a/pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
@@ -49,11 +49,8 @@ func (pl *ClusterResourcesLeastAllocated) Score(
 		return 0, framework.NewResult(framework.Error, err.Error())
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
-		allocatable[resource], requested[resource] = calculateResourceAllocatableRequest(su, cluster, resource)
-	}
+	resources := getRelevantResources(su)
+	allocatable, requested := getAllocatableAndRequested(su, cluster, resources)
 
 	var score, weightSum int64
 	// least allocated score favors cluster with fewer requested resources.
@@ -61,9 +58,15 @@ func (pl *ClusterResourcesLeastAllocated) Score(
 	// prioritizes based on the minimum of the average of the fraction of requested to capacity.
 	//
 	// Details:
-	// (cpu((capacity-sum(requested))*10/capacity) + memory((capacity-sum(requested))*10/capacity))/2
-	for resource, weight := range framework.DefaultRequestedRatioResources {
+	// (cpu((capacity-sum(requested))*100/capacity) * cpu_weight +
+	// memory((capacity-sum(requested))*100/capacity) * memory_weight) / (cpu_weight + memory_weight)
+	// Or with gpu
+	// (cpu((capacity-sum(requested))*100/capacity) * cpu_weight +
+	// memory((capacity-sum(requested))*100/capacity) * memory_weight +
+	// gpu((capacity-sum(requested))*100/capacity) * gpu_weight) / (cpu_weight + memory_weight + gpu_weight)
+	for _, resource := range resources {
 		resourceScore := leastRequestedScore(requested[resource], allocatable[resource])
+		weight := framework.DefaultRequestedRatioResources[resource]
 		score += resourceScore * weight
 		weightSum += weight
 	}
@@ -80,8 +83,8 @@ func (pl *ClusterResourcesLeastAllocated) ScoreExtensions() framework.ScoreExten
 	return nil
 }
 
-// The unused capacity is calculated on a scale of 0-10
-// 0 being the lowest priority and 10 being the highest.
+// The unused capacity is calculated on a scale of 0-100
+// 0 being the lowest priority and 100 being the highest.
 // The more unused resources the higher the score is.
 func leastRequestedScore(requested, capacity int64) int64 {
 	if capacity == 0 {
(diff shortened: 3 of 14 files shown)


Poor12:
/assign @gary-lgy @limhawjia @mrlihanbo 
No rush to review if you are on vacation. :)

pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -152,6 +154,20 @@ func schedulingUnitForFedObject(
 		schedulingUnit.MaxClusters = maxClustersOverride
 	}
 
+	gvk := schedulingUnit.GroupVersion.WithKind(schedulingUnit.Kind)
+	podSpec, err := fedpodutil.GetResourcePodSpec(fedObject, gvk)
+	// if we cannot recognize the type of resourceObj,
+	// do not return so that it will not block the entire scheduling process.
+	if err == nil {
+		resourceQuest := *framework.NewResource(resourceutil.GetPodResourceRequests(podSpec))
```
mrlihanbo, line 162: resourceRequest？

pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -152,6 +154,20 @@ func schedulingUnitForFedObject(
...
+		gpuResourceQuest := &framework.Resource{}
+		if resourceQuest.HasGivenScalarResource(framework.ResourceGPU) {
+			gpuResourceQuest.SetScalar(framework.ResourceGPU, resourceQuest.ScalarResources[framework.ResourceGPU])
+		}
+		// now we only consider the resource quest of gpu
```
mrlihanbo, line 167: request

pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -152,6 +154,20 @@ func schedulingUnitForFedObject(
 		schedulingUnit.MaxClusters = maxClustersOverride
 	}
 
+	gvk := schedulingUnit.GroupVersion.WithKind(schedulingUnit.Kind)
```
mrlihanbo, line 157:
```suggestion
	gvk := schedulingUnit.GroupVersion.WithKind(schedulingUnit.Kind)
	path, ok := fedpodutil.PodSpecPaths[gvk.GroupKind()]
	if !ok {
		return schedulingUnit, nil
	}
	
	podSpec, err := fedpodutil.GetPodSpec(fedObject, path)
	if err != nil {
		return nil, err
	}
	
	resourceQuest := *framework.NewResource(resourceutil.GetPodResourceRequests(podSpec))
	gpuResourceQuest := &framework.Resource{}
	if resourceQuest.HasGivenScalarResource(framework.ResourceGPU) {
		gpuResourceQuest.SetScalar(framework.ResourceGPU, resourceQuest.ScalarResources[framework.ResourceGPU])
	}
	// now we only consider the resource quest of gpu
	schedulingUnit.ResourceRequest = *gpuResourceQuest

	return schedulingUnit, nil
```

pkg/controllers/scheduler/schedulingunit.go
mrlihanbo, line 162: We obtain the gpu resources from requests, what if users specify GPU limits without specifying requests

mrlihanbo: https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/
image

pkg/controllers/scheduler/framework/util.go
```diff
@@ -59,7 +63,7 @@ const (
 
 // TODO(feature), make the RequestedRatioResources configable
 
-var DefaultRequestedRatioResources = ResourceToWeightMap{corev1.ResourceMemory: 1, corev1.ResourceCPU: 1}
+var DefaultRequestedRatioResources = ResourceToWeightMap{corev1.ResourceMemory: 1, corev1.ResourceCPU: 1, ResourceGPU: 4}
```
mrlihanbo, line 66: Should we need to explain why GPU ratio is 4?

pkg/controllers/scheduler/framework/plugins/clusterresources/most_allocated.go
```diff
@@ -49,19 +51,37 @@ func (pl *ClusterResourcesMostAllocated) Score(
 		return 0, framework.NewResult(framework.Error, err.Error())
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
+	resources := make([]corev1.ResourceName, 0, len(framework.DefaultRequestedRatioResources))
 	for resource := range framework.DefaultRequestedRatioResources {
+		switch resource {
```
mrlihanbo, line 56: We have the same code in three places, can we extract the code to a function?

Poor12: Done.

pkg/controllers/scheduler/framework/util.go
Poor12: Nice suggestion! Added some comments.

pkg/controllers/scheduler/schedulingunit.go
Poor12: img_v2_dff14454-79da-44fb-8fe0-52f78ccfa04g

pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -152,6 +154,24 @@ func schedulingUnitForFedObject(
...
+	_, ok := fedpodutil.PodSpecPaths[gvk.GroupKind()]
+	if !ok {
+		return schedulingUnit, nil
+	}
+	podSpec, err := fedpodutil.GetResourcePodSpec(fedObject, gvk)
+	if err != nil {
+		return nil, err
+	}
+
+	resourceRequest := *framework.NewResource(resourceutil.GetPodResourceRequests(podSpec))
```
JackZxj, line 167: You can use `getResourceRequest(typeConfig, fedObject)` here

JackZxj: And shall we move `getResourceRequest` to util.go?

Poor12: I'd like to move it to util.go because it can be used both in schedulingtriggers and scheduler.

pkg/controllers/scheduler/framework/util.go
```diff
@@ -481,3 +498,21 @@ func DefaultNormalizeScore(maxPriority int64, reverse bool, scores ClusterScoreL
 	}
 	return nil
 }
+
+func GetRelevantResources(su *SchedulingUnit) []corev1.ResourceName {
+	resources := make([]corev1.ResourceName, 0, len(DefaultRequestedRatioResources))
+	for resourceName := range DefaultRequestedRatioResources {
```
manmanhappy, line 504: Most of the resource names follow the the same rules.Do we need these kinds of judgement?

Poor12: This logic is to mainly to keep the logic untouched without GPU requests.
For example, if we only has cpu and memory and no modification is made, it will become `(cpu * 1 + memory * 1 + 0 * 4) / 6`. This would not meet our expectations.

pkg/controllers/scheduler/framework/plugins/clusterresources/most_allocated.go
```diff
@@ -49,19 +49,25 @@ func (pl *ClusterResourcesMostAllocated) Score(
 		return 0, framework.NewResult(framework.Error, err.Error())
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
+	resources := framework.GetRelevantResources(su)
```
manmanhappy, line 52: There are the same logic for getting variables like resources, requested and allocatable in balanced/least/most_allocated.go files.Can u extract them as a utility function?

pkg/controllers/scheduler/schedulingunit.go
Poor12: I made a mistake here. In the actual test, the Pod's spec will automatically fill in the request value, but the Deployment and so on will not. Since we get the request from the deployment's spec, we still need to deal with this special situation. @mrlihanbo @manmanhappy

pkg/controllers/scheduler/framework/plugins/clusterresources/most_allocated.go
Poor12: Done

mrlihanbo (approved):

pkg/util/resource/resource.go
```diff
@@ -45,6 +49,12 @@ func GetPodResourceRequests(podSpec *corev1.PodSpec) corev1.ResourceList {
 	reqs := make(corev1.ResourceList)
 
 	for _, container := range podSpec.Containers {
+		gpuLimits, hasGpuLimits := container.Resources.Limits[ResourceGPU]
```
gary-lgy, line 52: Could you add a comment here to explain the rationale?

Poor12: Done.

gary-lgy: This function is used for more than 1 purposes. Are we confident that this change will not affect the other users of this function?

Poor12: I have checked all references. There will be no other effects.

pkg/controllers/scheduler/framework/util.go
```diff
@@ -245,6 +250,18 @@ func (r *Resource) SetMaxResource(rl corev1.ResourceList) {
 	}
 }
 
+// HasGivenScalarResource checks if Resource has the given scalar resource.
+func (r *Resource) HasGivenScalarResource(name corev1.ResourceName) bool {
+	if r.ScalarResources == nil {
+		return false
+	}
+
+	if _, exists := r.ScalarResources[name]; exists {
```
gary-lgy, line 259: Why should one use this function instead of indexing `r.ScalarResources` directly?

Poor12: Mainly to avoid the situation where ScalarResources is nil.

pkg/controllers/scheduler/framework/util.go
```diff
@@ -245,6 +250,18 @@ func (r *Resource) SetMaxResource(rl corev1.ResourceList) {
 	}
 }
 
+// HasGivenScalarResource checks if Resource has the given scalar resource.
+func (r *Resource) HasGivenScalarResource(name corev1.ResourceName) bool {
```
gary-lgy, line 254:
```suggestion
func (r *Resource) HasScalarResource(name corev1.ResourceName) bool {
```

pkg/controllers/scheduler/framework/plugins/clusterresources/fit.go
```diff
@@ -173,3 +173,27 @@ func calculateResourceAllocatableRequest(
...
+			resources = append(resources, resourceName)
+		}
+	}
+
+	return resources
+}
+
+func getAllocatableAndRequested(su *framework.SchedulingUnit, cluster *fedcorev1a1.FederatedCluster,
+	resources []corev1.ResourceName,
+) (framework.ResourceToValueMap, framework.ResourceToValueMap) {
```
gary-lgy, line 191:
```suggestion
func getAllocatableAndRequested(
    su *framework.SchedulingUnit,
    cluster *fedcorev1a1.FederatedCluster,
	resources []corev1.ResourceName,
) (framework.ResourceToValueMap, framework.ResourceToValueMap) {
```

pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
```diff
@@ -49,21 +49,22 @@ func (pl *ClusterResourcesLeastAllocated) Score(
...
 
 	var score, weightSum int64
 	// least allocated score favors cluster with fewer requested resources.
 	// It calculates the percentage of memory and CPU requested by pods scheduled on the cluster, and
 	// prioritizes based on the minimum of the average of the fraction of requested to capacity.
 	//
 	// Details:
-	// (cpu((capacity-sum(requested))*10/capacity) + memory((capacity-sum(requested))*10/capacity))/2
-	for resource, weight := range framework.DefaultRequestedRatioResources {
+	// (cpu((capacity-sum(requested))*100/capacity) + memory((capacity-sum(requested))*100/capacity))/2
```
gary-lgy, line 61: This example assumes weight ratios of 1:1:4. It's possible that we might change the ratio in the code and forget to update these comments. Should we make it clear here that we're assuming 1:1:4?

Poor12: How about not using 4 but using `gpu weight` in comments?

pkg/controllers/scheduler/framework/plugins/clusterresources/most_allocated_test.go
```diff
@@ -75,6 +75,28 @@ func TestClusterResourcesMostAllocated(t *testing.T) {
 			},
 			name: "nothing scheduled, resources requested, differently sized machines",
 		},
+		{
+			// Cluster1 scores on 0-100 scale
+			// CPU Fraction: 3000 / 4000= 75%
+			// Memory Fraction: 5000 / 10000 = 50%
+			// GPU Fraction: 5000 / 10000 = 50%
```
gary-lgy, line 82: Could we add another test case where the GPU fractions differ (to show the effect of the 1:1:4 ratio).

pkg/controllers/scheduler/framework/plugins/clusterresources/balanced_allocation.go
```diff
@@ -47,32 +45,47 @@ func (pl *ClusterResourcesBalancedAllocation) Score(
...
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
-		allocatable[resource], requested[resource] = calculateResourceAllocatableRequest(su, cluster, resource)
+	resources := getRelevantResources(su)
+	allocatable, requested := getAllocatableAndRequested(su, cluster, resources)
+
+	resourceToFractions := make([]float64, 0, len(requested))
+	for i := range requested {
+		fraction := fractionOfCapacity(requested[i], allocatable[i])
+		// This to find a node which has most balanced CPU, memory and volume usage.
```
gary-lgy, line 61:
```suggestion
		// This to find a cluster which has most balanced resource usage.
```

pkg/controllers/scheduler/framework/plugins/clusterresources/balanced_allocation.go
```diff
@@ -47,32 +45,47 @@ func (pl *ClusterResourcesBalancedAllocation) Score(
...
 
-	// Upper and lower boundary of difference between cpuFraction and memoryFraction are -1 and 1
-	// respectively. Multiplying the absolute value of the difference by 10 scales the value to
-	// 0-10 with 0 representing well balanced allocation and 10 poorly balanced. Subtracting it from
-	// 10 leads to the score which also scales from 0 to 10 while 10 representing well balanced.
-	diff := math.Abs(cpuFraction - memoryFraction)
-	score := int64((1 - diff) * float64(framework.MaxClusterScore))
-	return score, framework.NewResult(framework.Success)
+	// STD (standard deviation) is always a positive value. 1-deviation lets the score to be higher for node which has least deviation and
+	// multiplying it with `MaxNodeScore` provides the scaling factor needed.
```
gary-lgy, line 87:
```suggestion
	// STD (standard deviation) is always a positive value. 1-deviation lets the score to be higher for cluster which has least deviation and
	// multiplying it with `MaxClusterScore` provides the scaling factor needed.
```

pkg/controllers/scheduler/framework/plugins/clusterresources/balanced_allocation.go
```diff
@@ -47,32 +45,47 @@ func (pl *ClusterResourcesBalancedAllocation) Score(
...
 	}
 
-	requested := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	allocatable := make(framework.ResourceToValueMap, len(framework.DefaultRequestedRatioResources))
-	for resource := range framework.DefaultRequestedRatioResources {
-		allocatable[resource], requested[resource] = calculateResourceAllocatableRequest(su, cluster, resource)
+	resources := getRelevantResources(su)
+	allocatable, requested := getAllocatableAndRequested(su, cluster, resources)
+
+	resourceToFractions := make([]float64, 0, len(requested))
```
gary-lgy, line 58: `resourceToFractions` sounds like a map; could we name this simply `fractions`?

gary-lgy (approved):

pkg/controllers/scheduler/framework/util.go
gary-lgy: Could we further explain why GPU is weighted 4? Simply "based on empirical experience" may not convince critical readers.

pkg/controllers/scheduler/framework/util.go
gary-lgy: Reading a nil map does not panic; only writing to a nil map panics

pkg/util/resource/resource.go
```diff
@@ -45,6 +49,16 @@ func GetPodResourceRequests(podSpec *corev1.PodSpec) corev1.ResourceList {
 	reqs := make(corev1.ResourceList)
 
 	for _, container := range podSpec.Containers {
+		// GPUs are only supposed to be specified in the limits section.
+		// Therefore, when only GPU is specified in the limits section,
+		// we need to obtain the resource request of GPU from limits.
+		// See details in https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/.
```
gary-lgy, line 55: Should we extend this logic to all resource types?

pkg/controllers/scheduler/framework/plugins/clusterresources/most_allocated_test.go
Poor12: Added.

pkg/util/resource/resource.go
Poor12: I think it'ok.

PR #228, opened 2023-09-06
chore: mark generated files as linguist-generated

SOF3:
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+zz_generated* linguist-generated=true
+pkg/client linguist-generated=true
+config/crds/*.yaml linguist-generated=true


.gitattributes
```diff
@@ -0,0 +1,3 @@
+zz_generated* linguist-generated=true
+pkg/client linguist-generated=true
+config/crds/*.yaml linguist-generated=true
```
gary-lgy, line 3: Will changes in these files still show up in the diff? We might still want to know that they changed without knowing what exactly has changed.

SOF3: They will, just collapsed by default

SOF3: Example: https://github.com/kubewharf/kelemetry/pull/109/files#diff-be72556c262b17baaa99cc522dd96991b063b2a9e5ee3af66fc20feb8768d00f
image

SOF3: In addition to collapsing in diff, another effect is that it will exclude generated files from code line stats, but idk if this will affect existing contributor line count

gary-lgy (approved):

PR #231, opened 2023-09-12
feat: add fitError when no available clusters

Poor12:
When there is no available clusters, scheduler will result in a `FitError`, and we can get the scheduling details from events.
```
 Warning  ScheduleFederatedObject  17m (x4 over 17m)     scheduler                  0/3 clusters are available: 1 cluster(s) didn't match cluster selector, 2 cluster(s) were not in placement list

```

### DIFF
--- a/pkg/controllers/scheduler/framework/plugins/apiresources/apiresources.go
+++ b/pkg/controllers/scheduler/framework/plugins/apiresources/apiresources.go
@@ -39,5 +39,5 @@ func (pl *APIResources) Filter(ctx context.Context, su *framework.SchedulingUnit
 			return framework.NewResult(framework.Success)
 		}
 	}
-	return framework.NewResult(framework.Unschedulable, "No matched group version kind.")
+	return framework.NewResult(framework.Unschedulable, "cluster(s) didn't support this APIVersion")
 }
--- a/pkg/controllers/scheduler/core/generic_scheduler.go
+++ b/pkg/controllers/scheduler/core/generic_scheduler.go
@@ -104,14 +104,17 @@ func (g *genericScheduler) Schedule(
 		return result, nil
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/controllers/scheduler/framework/plugins/apiresources/apiresources.go
```diff
@@ -39,5 +39,5 @@ func (pl *APIResources) Filter(ctx context.Context, su *framework.SchedulingUnit
 			return framework.NewResult(framework.Success)
 		}
 	}
-	return framework.NewResult(framework.Unschedulable, "No matched group version kind.")
+	return framework.NewResult(framework.Unschedulable, "cluster(s) didn't have the matched group version kind")
```
gary-lgy, line 42:
```suggestion
	return framework.NewResult(framework.Unschedulable, "cluster(s) didn't support this APIVersion")
```
Would this message be clearer to users?

gary-lgy:
Looks like our webhook plugins do not return a reason when a cluster is unschedulable. Should we open an issue to track it?

pkg/controllers/scheduler/framework/plugins/apiresources/apiresources.go
Poor12: Make senses.

Poor12:
> Looks like our webhook plugins do not return a reason when a cluster is unschedulable. Should we open an issue to track it?

Sure, maybe we can propose one right now, such as ("cluster(s) were filtered by webhookPlugin(%s)", p.name)

gary-lgy:
> > Looks like our webhook plugins do not return a reason when a cluster is unschedulable. Should we open an issue to track it?
> 
> Sure, maybe we can propose one right now, such as ("cluster(s) were filtered by webhookPlugin(%s)", p.name)

I was thinking more along the lines of letting webhooks return their own reasons. This would require extending our request and response definitions. For now, I think we can adopt your suggestion.

gary-lgy (approved):

PR #232, opened 2023-09-13
feat: support force delete for federated cluster

wy-lucky:
When attempting to delete a federated cluster, if it has the annotation **kubeadmiral.io/force-delete**, KubeAdmiral will forcefully delete the cluster.

### DIFF
--- a/pkg/controllers/federatedcluster/controller.go
+++ b/pkg/controllers/federatedcluster/controller.go
@@ -63,6 +63,9 @@ const (
 	EventReasonHandleTerminatingClusterBlocked = "HandleTerminatingClusterBlocked"
 )
 
+// annotationForcibleDelete on a federated cluster means that cluster should be forcible delete by KubeAdmiral.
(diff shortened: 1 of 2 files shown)


pkg/controllers/federatedcluster/controller.go
```diff
@@ -63,6 +63,9 @@ const (
 	EventReasonHandleTerminatingClusterBlocked = "HandleTerminatingClusterBlocked"
 )
 
+// annotationForceDelete on a federated cluster means that cluster should be force delete by KubeAdmiral.
+var annotationForceDelete = common.DefaultPrefix + "force-delete"
```
gary-lgy, line 67:
```suggestion
const AnnotationForceDelete = common.DefaultPrefix + "force-delete"
```
Please merge this with the const block above.

wy-lucky: done.

PR #233, opened 2023-09-15
feat(fcluster): do not collect resources related to virtual nodes.

wy-lucky:


cmd/controller-manager/app/options/options.go
```diff
@@ -140,7 +142,17 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		&o.PrometheusQuantiles,
 		"prometheus-quantiles",
 		map[string]string{"0.5": "0.01", "0.95": "0.01", "0.99": "0.002"},
-		"prometheus summary objective quantiles",
+		"Prometheus summary objective quantiles",
+	)
+
+	flags.StringToStringVar(
+		&federatedcluster.VirtualNodeLabels,
+		"virtual-node-labels",
```
limhawjia, line 150: should we make the name of the flag more general? this option can also be used to filter out other types of nodes as well

cmd/controller-manager/app/options/options.go
```diff
@@ -140,7 +142,17 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
...
+		map[string]string{
+			"type":                             "virtual-kubelet",
+			"node.kubernetes.io/instance-type": "eklet",
```
SOF3, line 153: This feels very vendor-specific. What about moving this to the FederatedCluster definition?

SOF3 (changes requested):

pkg/controllers/federatedcluster/util.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/kubewharf/kubeadmiral/pkg/util/resource"
 )
 
+var VirtualNodeLabels map[string]string
```
SOF3, line 27: Please don't use global variables to pass options directly. This is not unit testable.

cmd/controller-manager/app/options/options.go
wy-lucky: yes, we should. the final node collection consists of schedulable nodes. so how about using unschedulableNodeLabels here？

cmd/controller-manager/app/options/options.go
wy-lucky: @limhawjia suggests adding command line parameters to facilitate user configuration. I also think "node.kubernetes.io/instance-type": "eklet" very vendor-specific, maybe we can delete it.

limhawjia: let's keep `"type": "virtual-kubelet"` and remove `"node.kubernetes.io/instance-type": "eklet"`?

cmd/controller-manager/app/options/options.go
limhawjia: how about `--resource-aggregation-node-filter`? using the term "unschedulable" here might be confusing

pkg/controllers/federatedcluster/util.go
limhawjia: @wy-lucky you can refer to other controller-specific options such as `NSAutoPropExcludeRegexp` or `ClusterJoinTimeout`

cmd/controller-manager/app/options/options.go
SOF3: Let's remove both? vk is *supposed* to be schedulable; it is just inappropriate in the eks case.

wy-lucky: vk is supposed to be schedulable, but there's another problem that makes us should filter it.
If a cluster contains virtual nodes, federation will perceive a significant amount of available resources in that cluster during resource scheduling. This leads to a disproportionately high scheduling weight for that cluster type. Consequently, during dynamic weight scheduling, most replicas will be scheduled to this cluster.

pkg/controllers/federatedcluster/util.go
wy-lucky: please take a look.

cmd/controller-manager/app/options/options.go
wy-lucky: please take a look.

cmd/controller-manager/app/options/options.go
wy-lucky: After discussing with @SOF3 , we have concluded that this may not be a general requirement. so i will not add the default parameter here. cc @limhawjia

cmd/controller-manager/app/options/options.go
```diff
@@ -140,7 +142,14 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
 		&o.PrometheusQuantiles,
 		"prometheus-quantiles",
 		map[string]string{"0.5": "0.01", "0.95": "0.01", "0.99": "0.002"},
-		"prometheus summary objective quantiles",
+		"Prometheus summary objective quantiles",
+	)
+
+	flags.StringToStringVar(
```
SOF3, line 148: Why not accept a thing here and use [`labels.Parse`](https://pkg.go.dev/k8s.io/apimachinery@v0.28.2/pkg/labels#Parse)?

wy-lucky: label is a map structure, so it looks like string to string is more appropriate, just like the previous parameter PrometheusQuantiles.

pkg/controllers/federatedcluster/util_test.go
```diff
@@ -260,7 +260,8 @@ func Test_aggregateResources(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			allocatable, available := aggregateResources(tc.nodes, tc.pods)
```
limhawjia, line 263: let's add a test case for filtering nodes by label

cmd/controller-manager/app/util.go
```diff
@@ -192,5 +193,15 @@ func getComponentConfig(opts *options.Options) (*controllercontext.ComponentConf
 		componentConfig.NSAutoPropExcludeRegexp = nsAutoPropExcludeRegexp
 	}
 
+	var labelSelectors []labels.Selector
```
limhawjia, line 196: note the lint error here

pkg/controllers/federatedcluster/util_test.go
wy-lucky: ok

cmd/controller-manager/app/util.go
wy-lucky: fixed

SOF3 (approved):

cmd/controller-manager/app/options/options.go
```diff
@@ -140,7 +142,14 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string, disabl
...
 		map[string]string{"0.5": "0.01", "0.95": "0.01", "0.99": "0.002"},
-		"prometheus summary objective quantiles",
+		"Prometheus summary objective quantiles",
+	)
+
+	flags.StringArrayVar(
+		&o.ResourceAggregationNodeFilter,
+		"resource-aggregation-node-filter",
+		[]string{},
+		"The specific label signifies that the node doesn't require resource aggregation.",
```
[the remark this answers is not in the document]
wy-lucky: updated

limhawjia (approved):

PR #238, opened 2023-09-20
fix(fcluster): last transition time of cluster conditions

gary-lgy:
Last transition time should only be updated when there's a transition.

### DIFF
--- a/pkg/controllers/federatedcluster/util.go
+++ b/pkg/controllers/federatedcluster/util.go
@@ -55,25 +55,25 @@ func setClusterCondition(
 func getNewClusterOfflineCondition(
 	status corev1.ConditionStatus,
 	conditionTime metav1.Time,
+	oldCondition *fedcorev1a1.ClusterCondition,
 ) fedcorev1a1.ClusterCondition {
 	condition := fedcorev1a1.ClusterCondition{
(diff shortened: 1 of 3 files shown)


pkg/controllers/federatedcluster/util.go
```diff
@@ -54,33 +54,34 @@ func setClusterCondition(
...
+	oldCondition *fedcorev1a1.ClusterCondition,
 ) fedcorev1a1.ClusterCondition {
 	condition := fedcorev1a1.ClusterCondition{
 		Type:               fedcorev1a1.ClusterOffline,
+		Status:             status,
 		LastProbeTime:      conditionTime,
 		LastTransitionTime: conditionTime,
 	}
+	if oldCondition != nil && oldCondition.Status == status {
+		condition.LastTransitionTime = oldCondition.LastTransitionTime
```
limhawjia, line 66: how about we do this in `setClusterCondition` instead of repeating it for each `getNewXXXCondition` function definition?

gary-lgy: fixed, ptal cc. @mrlihanbo

mrlihanbo (approved):

PR #240, opened 2023-09-25
chore(policy): remove deprecated parameters

JackZxj:


pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -241,20 +241,13 @@ func getAutoMigrationInfo(fedObject fedcorev1a1.GenericFederatedObject) (*framew
 }
 
 func getIsStickyClusterFromPolicy(policy fedcorev1a1.GenericPropagationPolicy) bool {
```
limhawjia, line 243: should we rename the field inside `SchedulingUnit` to `ReschedulnigDisabled`? cc @gary-lgy

gary-lgy (approved):

PR #251, opened 2023-10-25
ci: integrate the codecov

qclc:
Add code coverage detection and ensure that **60%or higher of new code**  is covered with unit tests.

Currently we only focus on the coverage in the `./pkg` directory, and will ignore the following packages or files in this directory:
  - "pkg/apis"
  - "pkg/client"
  - "*_test.go"
  - "fake_*.go"

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,22 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2%
+    # https://docs.codecov.com/docs/commit-status#patch-status
+    patch:
+      default:
+        target: 60%
+        if_ci_failed: error
+# https://docs.codecov.com/docs/ignoring-paths
+ignore:
+  - "config"
+  - "docs"
+  - "examples"
+  - "hack"
+  - "pkg/apis"
+  - "pkg/client"
+  - "test"
+  - "*_test.go"
+  - "fake_*.go"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,6 @@ on:
       - v*
   pull_request:
 
-# TODO: coverage
-
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


SOF3:
Does the coverage data here include packages without any test files? See https://github.com/golang/go/issues/24570

I did a workaround in kelemetry because of this issue https://github.com/kubewharf/kelemetry/commit/26472190a8aa3ed42bb69bd38f0d3aa241216464

qclc:
> Does the coverage data here include packages without any test files? See [golang/go#24570](https://github.com/golang/go/issues/24570)

I noticed this problem and it seems to be solved by adding the `-coverpkg=./pkg/...` parameter to the `go test` command. When I experimented in the local repository, it seemed that the coverage would include all packages in the `./pkg` directory, even if there was no test file: https://github.com/qclc/kubeadmiral/pull/2#issuecomment-1777146050
According to experimental results, the old test report will not cover packages that do not contain single tests, but the new test report will cover all packages, even if there are no unit tests in them.

SOF3:
@qclc It seems the packages were included simply because your PR had diff in those packages. Maybe try with another commit that does not involve those packages? Would the total coverage decrease again?

qclc:
> @qclc It seems the packages were included simply because your PR had diff in those packages. Maybe try with another commit that does not involve those packages? Would the total coverage decrease again?

I experimented again and only modified the package containing the unit test. The coverage remained unchanged, which was also ~24% : https://app.codecov.io/gh/qclc/kubeadmiral/pull/4 . I also ran `go test` locally and observed the test report. After adding the `-coverpkg=./pkg/...` parameter, the test report will include all the packages in the `./pkg` directory. Otherwise, it won't. It seems that the test report after adding the `-coverpkg=./pkg/...` parameter is fully covered.

qclc:
Sorry to miss this comment on issue (https://github.com/golang/go/issues/24570#issuecomment-499863491) : using `-coverpkg` does not actually help if the package is never called in any test at all. `-coverpkg` is a cover-up that allows packages from a different package to provide coverage for this package.

The end result still requires some additional work, as @SOF3  did in repository kelemetry. Please take a review.

.codecov.yml
```diff
@@ -0,0 +1,22 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2%
+    # https://docs.codecov.com/docs/commit-status#patch-status
+    patch:
+      default:
+        target: 60%
```
SOF3, line 10: Should we really set 60% as target?

SOF3 (approved):

PR #252, opened 2023-10-25
[bugfix] fix the issue that rescheduling may mistakenly evict workload

Poor12:


pkg/controllers/scheduler/framework/plugins/clusterresources/fit.go
```diff
@@ -54,6 +54,14 @@ func (pl *ClusterResourcesFit) Filter(
 		return framework.NewResult(framework.Error, err.Error())
 	}
 
+	// When a workload has been scheduled to current cluster, the available resources of current cluster
+	// include the amount of resources used by Pods that have been scheduled to the cluster.
+	// When there are few available resources in the cluster, rescheduling may mistakenly
+	// evict workloads from the current cluster.
```
gary-lgy, line 60:
```suggestion
	// When a workload has been scheduled to the current cluster, the available resources of the current cluster
	// currently does not (but should) include the amount of resources requested by Pods of the current workload.
	// In the absence of ample resource buffer, rescheduling may mistakenly
	// evict the workload from the current cluster.
	// Disable this plugin for rescheduling as a temporary workaround.
```

gary-lgy (approved):

PR #255, opened 2023-10-31
feat(override): add image overrider

qclc:


pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
 	ClusterAffinity []ClusterSelectorTerm `json:"clusterAffinity,omitempty"`
 }
 
+// Overriders contains a list of override patch
+// The order in which the override patches take effect is:
+// - Image
+// - JsonPatch
 type Overriders struct {
+	Image []ImageOverrider `json:"image,omitempty"`
```
mrlihanbo, line 89: Should add comments and // +optional here

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
 	// JsonPatch specifies overriders in a syntax similar to RFC6902 JSON Patch.
 	// +optional
 	JsonPatch []JsonPatchOverrider `json:"jsonpatch,omitempty"`
 }
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set
```
mrlihanbo, line 96:
```suggestion
	// ContainerName is ignored when ImagePath is set.
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
 	ClusterAffinity []ClusterSelectorTerm `json:"clusterAffinity,omitempty"`
 }
 
+// Overriders contains a list of override patch
```
gary-lgy, line 84:
```suggestion
// Overriders contains a list of override patches.
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
 type Overriders struct {
+	Image []ImageOverrider `json:"image,omitempty"`
 	// JsonPatch specifies overriders in a syntax similar to RFC6902 JSON Patch.
 	// +optional
 	JsonPatch []JsonPatchOverrider `json:"jsonpatch,omitempty"`
 }
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set
+	// If null, the image override rule applies to all containers.
```
gary-lgy, line 97:
```suggestion
	// If nil or empty, the image override rule applies to all containers.
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
+	// ImagePath indicates the path of target image field.
+	// Defaults to empty, the system will automatically detect image fields if the resource type is
+	// Pod, CronJob, Deployment, StatefulSet, DaemonSet or Job
+	// If not empty, the image override rule applies to specify image
+	// +optional
+	ImagePath string `json:"imagePath,omitempty"`
+
+	// ImageComponent is part of image name.
+	// +kubebuilder:validation:Enum=registry;repository;tag;digest
+	// +required
```
gary-lgy, line 111: There's no such marker.

gary-lgy: See the full list of kubebuilder markers https://book.kubebuilder.io/reference/markers/crd-validation

gary-lgy: Additionally, the `OverridePolicy` struct already has the `+kubebuilder:valudation:Required` marker, and it seems like the marker is passed down recursively to the struct's fields. Thus all fields would be required by default and we only need to explicitly mark the optional fields.

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
+	Image []ImageOverrider `json:"image,omitempty"`
 	// JsonPatch specifies overriders in a syntax similar to RFC6902 JSON Patch.
 	// +optional
 	JsonPatch []JsonPatchOverrider `json:"jsonpatch,omitempty"`
 }
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set
+	// If null, the image override rule applies to all containers.
+	// If specified, it should match one container or initContainer name in pod template.
```
gary-lgy, line 98:
```suggestion
	// Otherwise, this override targets the specified container or init container in the pod template.
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
+	// +optional
+	ContainerName string `json:"containerName,omitempty"`
+
+	// ImagePath indicates the path of target image field.
+	// Defaults to empty, the system will automatically detect image fields if the resource type is
+	// Pod, CronJob, Deployment, StatefulSet, DaemonSet or Job
+	// If not empty, the image override rule applies to specify image
+	// +optional
+	ImagePath string `json:"imagePath,omitempty"`
```
gary-lgy, line 107:
```suggestion
	// ImagePath indicates the image path to target.
	// If empty, the system will automatically detect the image path if the resource type is
	// Pod, CronJob, Deployment, StatefulSet, DaemonSet or Job.
	// +optional
	ImagePath string `json:"imagePath,omitempty"`
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
+
+	// ImageComponent is part of image name.
```
gary-lgy, line 109:
```suggestion
	// ImageComponent is the part of the image to override.
```

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+	case OperatorAdd:
+		newTag = i.tag + value
+	case OperatorReplace:
+		newTag = value
+	case OperatorRemove:
+		newTag = ""
+	}
+
+	if operator != OperatorRemove && !utilreference.AnchoredTagRegexp.MatchString(newTag) {
+		return
```
gary-lgy, line 71: Should we return an error?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+		i.registry = value
+	case OperatorRemove:
+		i.registry = ""
+	}
+}
+
+func (i *Image) OperateRepository(operator, value string) {
+	switch operator {
+	case OperatorAdd:
+		i.repository += value
```
gary-lgy, line 51: For repository and registry, are there scenarios where concatenation would make sense? @mrlihanbo @limhawjia In jsonpatch, `add`ing value to an existing field would replace its value. Should we follow the same semantics?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	i.digest = newDigest
+}
+
+func (i *Image) String() string {
+	fullName := i.repository
+
+	if i.registry != "" {
+		fullName = strings.Join([]string{i.registry, i.repository}, "/")
```
gary-lgy, line 99: Simple `+=` might be more readable.

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+// ParseImage returns a Components of the given image.
+func ParseImage(fullName string) (*Image, error) {
+	ref, err := reference.Parse(fullName)
+	if err != nil {
+		return nil, err
+	}
+
+	image := &Image{}
+	if named, ok := ref.(reference.Named); ok {
+		image.registry, image.repository = SplitRegistryAndRepository(named.Name())
```
gary-lgy, line 122: What about `image.registry, image.repository = named.Domain(), named.Path()`?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+	if digested, ok := ref.(reference.Digested); ok {
+		image.digest = digested.Digest().String()
+	}
+
+	return image, nil
+}
+
+func SplitRegistryAndRepository(name string) (registry, repository string) {
+	i := strings.IndexRune(name, '/')
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
```
gary-lgy, line 138: Is there a reference for this logic?

pkg/apis/core/v1alpha1/types_overridepolicy.go
gary-lgy: Let's also give an example, since users might wonder whether the path should be dot or slash separated.

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+const (
+	pathSeparator   = "/"
+	OperatorAdd     = "add"
+	OperatorRemove  = "remove"
+	OperatorReplace = "replace"
+)
+
+var PodSpecPrefix = map[string]string{
```
gary-lgy, line 37: Can we reuse the definitions in https://github.com/gary-lgy/kubeadmiral/blob/60ac07e254541e7581da7c121079dcf08c77838b/pkg/util/pod/pod.go?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+
+	for i := 0; i < len(segments)-1; i++ {
+		segment := segments[i]
+		switch obj := currentObj[segment].(type) {
+		case map[string]interface{}:
+			currentObj = obj
+		case []interface{}:
+			index, err := strconv.ParseInt(segments[i+1], 10, 32)
+			if err != nil {
+				return "", fmt.Errorf("segment(%s) of path is not a number", segments[i+1])
```
gary-lgy, line 158: This error message can be more comprehensive, e.g.
` fmt.Errorf("%s requires the field to be an array, found %T instead", strings.Join(segments[:i+2], "/"), obj)`

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+			if int(index) >= len(obj) {
```
gary-lgy, line 160: Can index be negative?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+				return "", fmt.Errorf("slice index out of range")
+			}
+			currentObj = obj[index].(map[string]interface{})
```
gary-lgy, line 163: Is there a change this might panic?

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+			i++
+		default:
+			return "", fmt.Errorf("field(%s) of unstructured obj is not map[string]interface{} or []interface{}", segment)
```
gary-lgy, line 166: Same as above, adding the path prefix would be more helpful.

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+
+	value, ok := currentObj[segments[len(segments)-1]].(string)
+	if !ok {
+		return "", fmt.Errorf("convert value under path(%s) into string failed", path)
```
gary-lgy, line 172:
```suggestion
		return "", fmt.Errorf("imagePath %q does not point to a string field", path)
```

go.mod
```diff
@@ -56,6 +57,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/opencontainers/go-digest v1.0.0 // indirect
```
gary-lgy, line 60: Is this used anywhere?

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,46 @@ type TargetClusters struct {
...
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set
+	// If null, the image override rule applies to all containers.
+	// If specified, it should match one container or initContainer name in pod template.
+	// +optional
+	ContainerName string `json:"containerName,omitempty"`
+
+	// ImagePath indicates the path of target image field.
+	// Defaults to empty, the system will automatically detect image fields if the resource type is
```
limhawjia, line 103:
```suggestion
	// Defaults to empty, the system will automatically resolve image fields if the resource type is
```

pkg/apis/core/v1alpha1/types_overridepolicy.go
limhawjia: "... automatically detect ..." -> "... automatically resolve ..."

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+		registry, repository = name[:i], name[i+1:]
+	}
+	return
+}
+
+func ParseValueFromUnstructuredObj(input *unstructured.Unstructured, path string) (string, error) {
+	segments := strings.Split(strings.Trim(path, pathSeparator), pathSeparator)
+	currentObj := input.Object
+
+	for i := 0; i < len(segments)-1; i++ {
```
gary-lgy, line 150: Might want to refer to the implementation of `unstructured.NestedField`, which is more elegant imo. The main difference from the current implementation is to switch on `currentObj`'s type, instead of `currentObj[segment]`'s type. Thus, we don't need `segments[i+1]` and should be easier to reason.

pkg/controllers/override/imageparse.go
limhawjia: Concatenation feels like a very specific operation that may not be very applicable to registries and repositories. In the case of registries, we would be concatenating to the top-level domain which doesn't really make sense and in the case of repositories, it would only be applicable if users following a specific naming convention.

I feel that there may be a case for supporting more fine-grained string operations, but it may be better to have a more general operator for that, e.g. one that is regex-based or sed-based

pkg/controllers/override/imageparse.go
limhawjia: Yes. We should return an error and propagate to the user in the form of an event.

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if i.digest != "" {
+		fullName = strings.Join([]string{fullName, i.digest}, "@")
+	}
+
+	return fullName
+}
+
+// ParseImage returns a Components of the given image.
```
limhawjia, line 113:
```suggestion
// ParseImage returns an Image from the given string.
```

pkg/controllers/override/imageparse.go
limhawjia: The grammar can be found [here](https://github.com/distribution/reference/blob/49c28499d219290c3226822e9cfcd4ede6d75379/reference.go#L1). 

I think the `[Domain](https://github.com/distribution/reference/blob/49c28499d219290c3226822e9cfcd4ede6d75379/reference.go#L151)` and `[Path](https://github.com/distribution/reference/blob/49c28499d219290c3226822e9cfcd4ede6d75379/reference.go#L160)` functions provide the functionality we need here (correct me if I'm wrong). 

I would prefer if we use their implementation and not reinvent the wheel.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: I think an error should be returned here, otherwise the user will not be able to understand why the override does not take effect. The code here has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imagepatch.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated here to check if the index is a negative number.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

go.mod
qclc: This is automatically generated after `go mod tidy`. It may be a package that the `github.com/distribution/reference v0.5.0` library indirectly depends on.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
limhawjia: Is the path here supposed to be a JSON Patch [pointer](https://jsonpatch.com/#json-pointer)? If yes, is there an established package that we can use to do the parsing here?

limhawjia (commented):
Could we add a few simple unit tests for the functions in `imageparse.go` and `imagepatch.go`?

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,48 @@ type TargetClusters struct {
...
+	Image []ImageOverrider `json:"image,omitempty"`
 	// JsonPatch specifies overriders in a syntax similar to RFC6902 JSON Patch.
 	// +optional
 	JsonPatch []JsonPatchOverrider `json:"jsonpatch,omitempty"`
 }
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set.
+	// If empty, the image override rule applies to all containers.
+	// Otherwise, this override targets the specified container or init container in the pod template.
```
gary-lgy, line 100:
```suggestion
	// Otherwise, this override targets the specified container(s) or init container(s) in the pod template.
```

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,203 @@
...
+	imageOverriders []fedcorev1a1.ImageOverrider,
+) (fedcorev1a1.OverridePatches, error) {
+	// patchMap(<imagePath, imageValue>) is used to store the newest image value of each image path
+	patchMap := make(map[string]string)
+
+	// parse patches and store the new image values in patchMap
+	for i := range imageOverriders {
+		imageOverrider := imageOverriders[i]
+		if imageOverrider.ImagePath != "" {
+			if err := parsePatchesFromImagePath(fedObject, &imageOverrider, patchMap); err != nil {
```
gary-lgy, line 54:
```suggestion
		imageOverrider := &imageOverriders[i]
		if imageOverrider.ImagePath != "" {
			if err := parsePatchesFromImagePath(fedObject, imageOverrider, patchMap); err != nil {
```
This avoids an unnecessary copy, assuming `parsePatchesFromImagePath` and `parsePatchesFromWorkload` do not mutate `imageOverrider`.

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,203 @@
...
+		imageValue, err = ParseValueFromUnstructuredObj(sourceObj, imageOverrider.ImagePath)
+		if err != nil {
+			return fmt.Errorf("failed to parse image value from unstructured obj: %w", err)
+		}
+	}
+
+	// apply image override
+	newImageValue, err := applyImageOverride(imageOverrider.ImagePath, imageValue, imageOverrider)
+	if err != nil {
+		return fmt.Errorf("failed apply image override: %w", err)
```
gary-lgy, line 107:
```suggestion
		return fmt.Errorf("failed to apply image override: %w", err)
```

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,181 @@
...
+	i := strings.IndexRune(name, '/')
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
+		registry, repository = "", name
+	} else {
+		registry, repository = name[:i], name[i+1:]
+	}
+	return
+}
+
+func ParseValueFromUnstructuredObj(input *unstructured.Unstructured, path string) (string, error) {
```
gary-lgy, line 148:
```suggestion
func GetStringFromUnstructuredObj(input *unstructured.Unstructured, path string) (string, error) {
```

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// convert user-specified containerNames into set
+	specifiedContainers := sets.New[string]()
+	for _, name := range imageOverrider.ContainerNames {
+		if name != "" {
+			specifiedContainers.Insert(name)
+		}
+	}
+
+	for index, container := range podSpec.Containers {
```
gary-lgy, line 140: Did we forget to include init containers?

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,203 @@
...
+			patchMap[imagePath] = newImageValue
+		}
+	}
+
+	return nil
+}
+
+// applyImageOverride applies override to oldImage to generate a new value
+func applyImageOverride(imagePath, oldImage string, imageOverrider *fedcorev1a1.ImageOverrider) (string, error) {
+	if !strings.HasPrefix(imagePath, pathSeparator) {
```
gary-lgy, line 165: This check seems a bit out-of-place. Any reason it's performed here instead of earlier?

qclc: yes, currently we only need to verify the image path specified by user. The code here has been placed in the `parsePatchesFromImagePath` function.

pkg/controllers/override/imagepatch.go
qclc: The relevant code has been added.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imagepatch.go
qclc: The code has been updated.

pkg/controllers/override/imagepatch.go
qclc: The code has been updated.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

qclc:
> Could we add a few simple unit tests for the functions in `imageparse.go` and `imagepatch.go`?

Unit tests for image override has been added.

pkg/controllers/override/imagepatch.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: the reference is here: https://github.com/distribution/distribution/blob/v2.7.1/reference/normalize.go#L62

qclc: The reason for using reference here is due to this discussion: https://stackoverflow.com/a/37867949

If we use `image.registry, image.repository = named.Domain(), named.Path()` for parsing, the following problems will occur: `ealen/echo-server` is resolved to `registry: "ealen", repository : "echo-server"`.

Using the function in the reference ensures that the above image is parsed into the following content: `registry: "", repository: "ealen/echo-server"`

pkg/controllers/override/imageparse.go
qclc: The logic of the Add operation in ImageOverrider has been modified to be consistent with jsonPatch semantics. If the original value is not empty, it will be replaced instead of extended.

pkg/controllers/override/imageparse.go
gary-lgy: Please add this explanation to the comments so future maintainers would know about the context and be careful with this logic.

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,176 @@
...
+					field,
+				)
+			}
+			if index < 0 || int(index) >= len(obj) {
+				return "", fmt.Errorf("index(%d) is negative or out of range", index)
+			}
+			currentObj = obj[index]
+		default:
+			return "", fmt.Errorf(
+				"%s accessor error: unkowen type %T, expected map[string]interface{} or []interface{}",
```
gary-lgy, line 164:
```suggestion
				"%s accessor error: unknown type %T, expected map[string]interface{} or []interface{}",
```

pkg/controllers/override/imageparse.go
```diff
@@ -0,0 +1,176 @@
...
+				strings.Join(fields[:i+1], "/"),
```
gary-lgy, line 165:
```suggestion
				strings.Join(fields[:i], "/"),
```
Are we off by 1 here?

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,253 @@
...
+	// convert user-specified containerNames into set
+	specifiedContainers := sets.New[string]()
+	for _, name := range imageOverrider.ContainerNames {
+		if name != "" {
+			specifiedContainers.Insert(name)
+		}
+	}
+
+	// process the init containers
+	for index, container := range podSpec.InitContainers {
```
gary-lgy, line 150:
```suggestion
    for _, containers := range [][]corev1.Container{podSpec.InitContainers, podSpec.Containers} {
		for index, container := range containers {
```

Duplication is unhealthy, and can be avoided like this.

pkg/controllers/override/imagepatch.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+		default:
+			return "", fmt.Errorf("unsupported image component(%s)", component)
+		}
+	}
+
+	return newImage.String(), nil
+}
+
+func generateImagePathOfPodContainer(gvk schema.GroupVersionKind, index int) (string, error) {
```
gary-lgy, line 237: Please make "containers" and "initContainers" a parameter and merge the two functions into one.

qclc: The code has been updated.

pkg/controllers/override/imagepatch.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: yes, this should be `i`, the code has been updated.

pkg/controllers/override/imageparse.go
qclc: The code has been updated.

pkg/controllers/override/imageparse.go
qclc: The comments has been added.

pkg/apis/core/v1alpha1/types_overridepolicy.go
```diff
@@ -81,12 +81,55 @@ type TargetClusters struct {
...
+	// Image specifies the overriders that applies to the image.
+	// +optional
+	Image []ImageOverrider `json:"image,omitempty"`
 	// JsonPatch specifies overriders in a syntax similar to RFC6902 JSON Patch.
 	// +optional
 	JsonPatch []JsonPatchOverrider `json:"jsonpatch,omitempty"`
 }
 
+type ImageOverrider struct {
+	// ContainerName is ignored when ImagePath is set.
```
mrlihanbo, line 98: ContainerNames?

pkg/controllers/override/imageparse.go
qclc: The logic of the Replace operation in ImageOverrider has been modified to be consistent with jsonPatch semantics. If the original value is empty, it will be refused.

pkg/controllers/override/imageparse.go
qclc: The comment has been added.

pkg/apis/core/v1alpha1/types_overridepolicy.go
qclc: yes, the code has been updated.

gary-lgy (approved):

mrlihanbo (approved):

PR #256, opened 2023-11-03
fix(overridepolicy-controller): fix using deepEqual incorrectly 

Poor12:
Now we use the following statement to indicate whether overrides are updated.
```
needsUpdate := !equality.Semantic.DeepEqual(overrides, currentOverrides)
```
But overrides's type is `overridesMap`, whereas `currentOverrides` is a slice. So needsUpdate is always true which may lead to some useless updates.

BTW, I discovered that the resource had this event when I created the resource but did not associate the overrides policy which made me feel strange.
```
updated overrides according to override polic(ies)
```
After I deep into the codes, I found this mainly happens when fedobject needs to be updated. It has two situations:
* override policy updates
* pending annotation updates, e.g. we found no ops and removed the override-policy from pending controllers.

In the latter situation, event information is inaccurate. What do you think? /cc @gary-lgy @limhawjia @mrlihanbo

### DIFF
--- a/pkg/controllers/override/overridepolicy_controller.go
+++ b/pkg/controllers/override/overridepolicy_controller.go
@@ -343,7 +343,7 @@ func (c *Controller) reconcile(ctx context.Context, qualifiedName common.Qualifi
 		return worker.StatusError
 	}
 
-	var overrides overridesMap
+	var overrides, currentOverrides overridesMap
 	// Apply overrides from each policy in order
 	for _, policy := range policies {
 		newOverrides, err := parseOverrides(policy, placedClusters, fedObject)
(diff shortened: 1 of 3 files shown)


gary-lgy (approved):

gary-lgy:
> BTW, I discovered that the resource had this event when I created the resource but did not associate the overrides policy which made me feel strange.
> 
> ```
> updated overrides according to override polic(ies)
> ```
> 
> After I deep into the codes, I found this mainly happens when fedobject needs to be updated. It has two situations:
> 
> * override policy updates
> * pending annotation updates, e.g. we found no ops and removed the override-policy from pending controllers.
> 
> In the latter situation, event information is inaccurate. What do you think? /cc @gary-lgy @limhawjia @mrlihanbo

Agree that in the second situation the event is misleading. Would you like to fix it?

Poor12:
> > BTW, I discovered that the resource had this event when I created the resource but did not associate the overrides policy which made me feel strange.
> > ```
> > updated overrides according to override polic(ies)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After I deep into the codes, I found this mainly happens when fedobject needs to be updated. It has two situations:
> > 
> > * override policy updates
> > * pending annotation updates, e.g. we found no ops and removed the override-policy from pending controllers.
> > 
> > In the latter situation, event information is inaccurate. What do you think? /cc @gary-lgy @limhawjia @mrlihanbo
> 
> Agree that in the second situation the event is misleading. Would you like to fix it?

I'd like to fix it later.

mrlihanbo (approved):

PR #263, opened 2023-11-09
fix(scheduler): prevents scheduling new replicas to terminating clusters

JackZxj:
The current scheduler does not filter clusters being deleted because it may be necessary to retain resources in member clusters. However, this situation resulted in:
1. One member cluster is stuck in deletion
2. Workload is distributed using Divide mode
3. The deleting cluster has a high weight after `filter` and `score`.

**The scheduler may schedule new replicas to this terminating cluster.** However the sync controller will not dispatch resources to the terminating clusters. This will cause the workload to fail to scale up in federation.

This PR ensures that the terminating cluster will not scale up by setting `MaxReplicas` for the terminating cluster. Specifically:
* If the cluster being deleted has already been assigned replicas, the MaxReplicas is the currently assigned value.
* Otherwise, the MaxReplicas will be set to 0 to avoid creating new replicas in the terminating cluster.

When the `MaxClusters` exists at the same time, the scheduler will also give priority to non-terminated clusters when scheduling is triggered, specifically:
* If there are non-terminated clusters larger than MaxClusters, scheduling will be prioritized in these non-terminated clusters.
* When the number of non-terminating clusters is less than MaxClusters, the terminating cluster will still be selected by the scheduler. At this time, the number of replicas that can be scheduled will not more then the currently assigned value or be 0

### DIFF
--- a/pkg/controllers/scheduler/framework/plugins/maxcluster/max_cluster_test.go
+++ b/pkg/controllers/scheduler/framework/plugins/maxcluster/max_cluster_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"reflect"
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
@@ -28,12 +29,16 @@ import (
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework"
 )
 
-func makeCluster(clusterName string) *fedcorev1a1.FederatedCluster {
-	return &fedcorev1a1.FederatedCluster{
+func makeCluster(clusterName string, terminating bool) *fedcorev1a1.FederatedCluster {
+	cluster := &fedcorev1a1.FederatedCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: clusterName,
 		},
 	}
+	if terminating {
+		cluster.DeletionTimestamp = &metav1.Time{Time: time.Now()}
+	}
+	return cluster
 }
 
 func newIntP64(in int64) *int64 {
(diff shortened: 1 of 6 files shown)


mrlihanbo (approved):

limhawjia (approved):

pkg/controllers/scheduler/framework/plugins/maxcluster/max_cluster_test.go
```diff
@@ -28,12 +29,16 @@ import (
...
-func makeCluster(clusterName string) *fedcorev1a1.FederatedCluster {
-	return &fedcorev1a1.FederatedCluster{
+func makeCluster(clusterName string, terminating bool) *fedcorev1a1.FederatedCluster {
+	cluster := &fedcorev1a1.FederatedCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: clusterName,
 		},
 	}
+	if terminating {
+		cluster.DeletionTimestamp = &metav1.Time{Time: time.Now()}
```
gary-lgy, line 39: nit (no need to change it for this PR): could simplify this to `metav1.Now`

pkg/controllers/scheduler/schedulingunit.go
```diff
@@ -609,6 +611,26 @@ func getMaxReplicasFromObject(object fedcorev1a1.GenericFederatedObject) (map[st
 	return maxReplicas, true
 }
 
+// updateMaxReplicasForTerminatingCluster prevents scheduling new replicas to terminating clusters
+func updateMaxReplicasForTerminatingCluster(
+	su *framework.SchedulingUnit,
+	clusters []*fedcorev1a1.FederatedCluster,
+) {
+	for _, cluster := range clusters {
```
gary-lgy, line 619: Should we run this regardless of scheduling mode?

gary-lgy (approved):

PR #265, opened 2023-11-10
feat: support custom migration

Poor12:
To support custom migration we define three types of annotations:
* `kubeadmiral.io/migration-info` on source object: User-entered migration strategy.
* `kubeadmiral.io/latest-migration-info` on source object: The current migration strategy in effect, read-only.

Test Results:
1. WorkloadMigration:
Before migration:
image
Add migration info anno:
image

After migration:
SourceObj:
image
FedObj:
image

2. ReplicasMigration:
Before migration:
image
Add migration info anno:
image

After migration:
SourceObj:
image
FedObj:
image

### DIFF
--- a/pkg/apis/core/v1alpha1/types_federatedtypeconfig.go
+++ b/pkg/apis/core/v1alpha1/types_federatedtypeconfig.go
@@ -53,8 +53,9 @@ type FederatedTypeConfigSpec struct {
 	// The API resource type to be federated.
 	SourceType APIResource `json:"sourceType"`
 
-	// Configuration for StatusAggregation. If left empty, the StatusAggregation feature will be disabled.
+	// Configuration for StatusAggregation.
 	// +optional
+	// +kubebuilder:default:={enabled:true}
 	StatusAggregation *StatusAggregationConfig `json:"statusAggregation,omitempty"`
 	// Configuration for StatusCollection. If left empty, the StatusCollection feature will be disabled.
 	// +optional
@@ -115,6 +116,7 @@ type StatusCollectionConfig struct {
 // StatusAggregationConfig defines the configurations for the StatusAggregation feature.
 type StatusAggregationConfig struct {
 	// Whether or not to enable status aggregation.
+	// +kubebuilder:default:=true
 	Enabled bool `json:"enabled"`
 }
 
--- a/pkg/controllers/common/constants.go
+++ b/pkg/controllers/common/constants.go
@@ -116,6 +116,11 @@ const (
 	TemplateGeneratorMergePatchAnnotation = FederateControllerPrefix + "template-generator-merge-patch"
 
 	LatestReplicasetDigestsAnnotation = DefaultPrefix + "latest-replicaset-digests"
+
+	// MigrationConfigurationAnnotation contains custom migration configuration from users.
+	MigrationConfigurationAnnotation = DefaultPrefix + "migration-configuration"
+	// AppliedMigrationConfigurationAnnotation contains the applied custom migration configuration.
+	AppliedMigrationConfigurationAnnotation = DefaultPrefix + "applied-migration-configuration"
 )
 
 // PropagatedAnnotationKeys and PropagatedLabelKeys are used to store the keys of annotations and labels that are present
--- a/pkg/controllers/scheduler/constants.go
+++ b/pkg/controllers/scheduler/constants.go
@@ -41,10 +41,11 @@ const (
 
 	DefaultSchedulingMode = fedcorev1a1.SchedulingModeDuplicate
 
-	EventReasonScheduleFederatedObject   = "ScheduleFederatedObject"
-	EventReasonInvalidFollowsObject      = "InvalidFollowsObject"
-	EventReasonWebhookConfigurationError = "WebhookConfigurationError"
-	EventReasonWebhookRegistered         = "WebhookRegistered"
+	EventReasonScheduleFederatedObject     = "ScheduleFederatedObject"
+	EventReasonInvalidFollowsObject        = "InvalidFollowsObject"
+	EventReasonWebhookConfigurationError   = "WebhookConfigurationError"
+	EventReasonWebhookRegistered           = "WebhookRegistered"
+	EventReasonParseMigrationConfiguration = "ParseMigrationConfiguration"
 
 	SchedulingTriggersAnnotation        = common.DefaultPrefix + "scheduling-triggers"
 	SchedulingDeferredReasonsAnnotation = common.DefaultPrefix + "scheduling-deferred-reasons"
--- a/pkg/controllers/scheduler/framework/plugins/clusterevicted/cluster_evicted.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterevicted/cluster_evicted.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2023 The KubeAdmiral Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterevicted
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	fedcorev1a1 "github.com/kubewharf/kubeadmiral/pkg/apis/core/v1alpha1"
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework"
+	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework/plugins/names"
+)
+
+const (
+	// ErrReason for cluster evicted not matching.
+	ErrReason = "cluster(s) were evicted due to custom migration configuration"
+)
+
+type ClusterEvicted struct{}
+
+func NewClusterEvicted(_ framework.Handle) (framework.Plugin, error) {
+	return &ClusterEvicted{}, nil
+}
+
+func (pl *ClusterEvicted) Name() string {
+	return names.ClusterEvicted
+}
+
+// Filter invoked at the filter extension point.
+func (pl *ClusterEvicted) Filter(
+	ctx context.Context,
+	su *framework.SchedulingUnit,
+	cluster *fedcorev1a1.FederatedCluster,
+) *framework.Result {
+	err := framework.PreCheck(ctx, su, cluster)
+	if err != nil {
+		return framework.NewResult(framework.Error, err.Error())
+	}
+
+	if su.CustomMigration.Info != nil {
+		unavailableClusters := su.CustomMigration.Info.UnavailableClusters
+		for _, unavailableCluster := range unavailableClusters {
+			if cluster.Name == unavailableCluster.Cluster && metav1.Now().Time.Before(unavailableCluster.ValidUntil.Time) {
+				return framework.NewResult(framework.Unschedulable, ErrReason)
+			}
+		}
+	}
+
+	return framework.NewResult(framework.Success)
+}
--- a/pkg/controllers/scheduler/framework/plugins/rsp/rsp.go
+++ b/pkg/controllers/scheduler/framework/plugins/rsp/rsp.go
@@ -153,6 +153,11 @@ func (pl *ClusterCapacityWeight) ReplicaScheduling(
 		}
 	}
 
+	limitedCapacity := map[string]int64{}
+	if su.CustomMigration.Info != nil && su.CustomMigration.Info.LimitedCapacity != nil {
+		limitedCapacity = su.CustomMigration.Info.LimitedCapacity
+	}
+
 	scheduleResult, overflow, err := planner.Plan(
 		&planner.ReplicaSchedulingPreference{
 			Clusters: clusterPreferences,
@@ -161,6 +166,7 @@ func (pl *ClusterCapacityWeight) ReplicaScheduling(
 		ExtractClusterNames(clusters),
 		currentReplicas,
 		estimatedCapacity,
+		limitedCapacity,
 		su.Key(),
 		su.AvoidDisruption,
 		keepUnschedulableReplicas,
--- a/pkg/controllers/scheduler/planner/planner_test.go
+++ b/pkg/controllers/scheduler/planner/planner_test.go
@@ -29,11 +29,12 @@ import (
 )
 
 type testCase struct {
-	rsp      map[string]ClusterPreferences
-	replicas int64
-	clusters []string
-	existing map[string]int64
-	capacity map[string]int64
+	rsp             map[string]ClusterPreferences
+	replicas        int64
+	clusters        []string
+	existing        map[string]int64
+	capacity        map[string]int64
+	limitedCapacity map[string]int64
 }
 
 type expectedResult struct {
@@ -75,7 +76,7 @@ func doCheck(
 				Clusters: tc.rsp,
 			},
 			tc.replicas, tc.clusters,
-			existing, estimatedCapacity, "",
+			existing, estimatedCapacity, tc.limitedCapacity, "",
 			avoidDisruption, keepUnschedulableReplicas,
 		)
 
@@ -426,6 +427,38 @@ func TestWithExisting(t *testing.T) {
 		[2]*expectedResult{
 			{plan: map[string]int64{"A": 25, "B": 25}},
 			{plan: map[string]int64{"A": 30, "B": 20}},
+		},
+	)
+
+	doCheckWithExisting(t,
+		&testCase{
+			rsp: map[string]ClusterPreferences{
+				"*": {Weight: 1},
+			},
+			replicas:        50,
+			clusters:        []string{"A", "B"},
+			existing:        map[string]int64{"A": 30},
+			limitedCapacity: map[string]int64{"A": 0},
+		},
+		[2]*expectedResult{
+			{plan: map[string]int64{"A": 0, "B": 50}},
+			{plan: map[string]int64{"A": 0, "B": 50}},
+		},
+	)
+
+	doCheckWithExisting(t,
+		&testCase{
+			rsp: map[string]ClusterPreferences{
+				"*": {Weight: 1},
+			},
+			replicas:        50,
+			clusters:        []string{"A", "B"},
+			existing:        map[string]int64{"A": 30},
+			limitedCapacity: map[string]int64{"A": 10},
+		},
+		[2]*expectedResult{
+			{plan: map[string]int64{"A": 10, "B": 40}},
+			{plan: map[string]int64{"A": 10, "B": 40}},
 		},
 	)
 
--- a/pkg/controllers/scheduler/scheduler.go
+++ b/pkg/controllers/scheduler/scheduler.go
@@ -361,6 +361,37 @@ func (s *Scheduler) prepareToSchedule(
 ) {
 	logger := klog.FromContext(ctx)
 
+	// update applied migration config
+	// We must update migration configuration before the checking scheduling triggers because
+	// the former is part of the latter.
+
+	existingMigrationConfigBytes, migrationConfigExists := fedObject.GetAnnotations()[common.MigrationConfigurationAnnotation]
+	_, customMigrationInfoExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
+	if !migrationConfigExists {
+		if customMigrationInfoExists {
+			annotations := fedObject.GetAnnotations()
+			delete(annotations, common.AppliedMigrationConfigurationAnnotation)
+			fedObject.SetAnnotations(annotations)
+		}
+	} else {
+		customMigrationInfoBytes, err := getCustomMigrationInfoBytes(existingMigrationConfigBytes, fedObject, ftc)
+		if err != nil {
+			logger.Error(err, "Failed to parse custom migration configuration")
+			s.eventRecorder.Eventf(
+				fedObject,
+				corev1.EventTypeWarning,
+				EventReasonParseMigrationConfiguration,
+				"Failed to parse custom migration configuration: %v",
+				err,
+			)
+		} else {
+			annotations := fedObject.GetAnnotations()
+			annotations[common.AppliedMigrationConfigurationAnnotation] = customMigrationInfoBytes
+			fedObject.SetAnnotations(annotations)
+			logger.V(3).Info("Successfully parsed custom migration configuration")
+		}
+	}
+
 	// check pending controllers
 
 	if ok, err := pendingcontrollers.ControllerDependenciesFulfilled(fedObject, PrefixedGlobalSchedulerName); err != nil {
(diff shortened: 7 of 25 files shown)


pkg/controllers/common/constants.go
```diff
@@ -116,6 +116,13 @@ const (
 	TemplateGeneratorMergePatchAnnotation = FederateControllerPrefix + "template-generator-merge-patch"
 
 	LatestReplicasetDigestsAnnotation = DefaultPrefix + "latest-replicaset-digests"
+
+	// MigrationInfoAnnotation contains custom migration information from users on the source object.
+	MigrationInfoAnnotation = DefaultPrefix + "migration-info"
+	// LatestMigrationInfoAnnotation contains latest custom migration information from users on the source object.
+	LatestMigrationInfoAnnotation = DefaultPrefix + "latest-migration-info"
+	// CustomMigrationInfoAnnotation contains custom migration information from users on the federated object.
+	CustomMigrationInfoAnnotation = DefaultPrefix + "custom-migration-info"
```
Poor12, line 125: CustomMigrationInfoAnnotation is an internal annotation. Use InternalPrefix instead.

mrlihanbo, line 125:
```suggestion
	CustomMigrationInfoAnnotation = InternalPrefix + "custom-migration-info"
```

pkg/controllers/common/constants.go
```diff
@@ -116,6 +116,13 @@ const (
 	TemplateGeneratorMergePatchAnnotation = FederateControllerPrefix + "template-generator-merge-patch"
 
 	LatestReplicasetDigestsAnnotation = DefaultPrefix + "latest-replicaset-digests"
+
+	// MigrationInfoAnnotation contains custom migration information from users on the source object.
+	MigrationInfoAnnotation = DefaultPrefix + "migration-info"
+	// LatestMigrationInfoAnnotation contains latest custom migration information from users on the source object.
```
mrlihanbo, line 122: What is the difference between MigrationInfoAnnotation and LatestMigrationInfoAnnotation

pkg/controllers/federate/controller.go
```diff
@@ -519,6 +520,14 @@ func (c *FederateController) handleCreateFederatedObject(
 		return fmt.Errorf("failed to set pending controllers on federated object: %w", err)
 	}
 
+	if needsUpdate {
```
gary-lgy, line 523: Migration config has not been applied at this stage, it can be misleading to populate the annotation so soon. 
Consider adding the annotation after the scheduler has processed the migration config.

pkg/controllers/federate/controller.go
```diff
@@ -541,10 +550,19 @@ func (c *FederateController) handleExistingFederatedObject(
 	logger := klog.FromContext(ctx)
 
 	logger.V(3).Info("Checking if federated object needs update")
-	needsUpdate, err := updateFederatedObjectForSourceObject(ftc, sourceObject, fedObject)
+	needsUpdate, sourceObjNeedsUpdate, err := updateFederatedObjectForSourceObject(ctx, ftc, sourceObject, fedObject, c.eventRecorder)
 	if err != nil {
 		return false, fmt.Errorf("failed to check if federated object needs update: %w", err)
 	}
+
+	if sourceObjNeedsUpdate {
```
gary-lgy, line 558: If I understood correctly, with this PR, the new annotations are processed as follows:
1. User writes `kubeadmiral.io/migration-info` on the source object
2. federate controller does the following when converting the source object to the federated object:
	2.1. parses it and validates it
    2.2. converts it to `kubeadmiral.io/custom-migration-info` and writes it to the fedobject
    2.3. converts it to `kubeadmiral.io/latest-migration-info` and writes it to the source object
3. scheduler reads `kubeadmiral.io/custom-migration-info` and schedules accordingly

Migration/Eviction should be in the scheduler's domain, and letting federate controller process it breaks abstraction and doesn't really fit well into the current architecture. Without clear boundaries, controllers become hard to maintain and understand.

Meanwhile, step 2.3 doesn't synchronize with step 3, so the reported `latest-migration-info` may not even have taken effect.

I propose the following solution to the above problems:
1. We already have the mechanism to copy source annotations to the fed object (see `classifyAnnotations`). We can use it to copy `kubeadmiral.io/migration-info` to the fedObject.
2. Leave the validation and processing to the scheduler. Once the scheduler has rescheduled based on the new migration config, it writes `kubeadmiral.io/latest-migration-info` to the fedObject at the same time it updates the fedObject placement.
3. `sourcefeedback` in federate controller copies `kubeadmiral.io/latest-migration-info` to the source object.

cc @limhawjia @mrlihanbo Please share your opinions too.

gary-lgy (changes requested):

pkg/controllers/federate/controller.go
mrlihanbo: It is a little stranger to put the `sourcefeedback` in federate controller

pkg/controllers/scheduler/framework/plugins/clusterevicted/cluster_evicted.go
```diff
@@ -0,0 +1,65 @@
...
+	su *framework.SchedulingUnit,
+	cluster *fedcorev1a1.FederatedCluster,
+) *framework.Result {
+	err := framework.PreCheck(ctx, su, cluster)
+	if err != nil {
+		return framework.NewResult(framework.Error, err.Error())
+	}
+
+	if su.CustomMigration != nil && su.CustomMigration.Info != nil {
+		unavailableClusters := su.CustomMigration.Info.UnavailableClusters
```
mrlihanbo, line 56: evictingClusters may be more suitable? /cc @gary-lgy

pkg/apis/core/v1alpha1/types_federatedtypeconfig.go
```diff
@@ -53,9 +53,9 @@ type FederatedTypeConfigSpec struct {
 	// The API resource type to be federated.
 	SourceType APIResource `json:"sourceType"`
 
-	// Configuration for StatusAggregation. If left empty, the StatusAggregation feature will be disabled.
-	// +optional
-	StatusAggregation *StatusAggregationConfig `json:"statusAggregation,omitempty"`
+	// Configuration for StatusAggregation.
+	// +required
+	StatusAggregation StatusAggregationConfig `json:"statusAggregation"`
```
REestrella, line 58: StatusAggregation is default true, unless we turn it off explicitly. To make API more clearer, I prefer to make this change.

Poor12, line 58: StatusAggregation is default true, unless we turn it off explicitly. To make API clearer, I prefer to make this change.

pkg/controllers/scheduler/planner/planner_test.go
```diff
@@ -429,6 +430,22 @@ func TestWithExisting(t *testing.T) {
 		},
 	)
 
+	doCheckWithExisting(t,
+		&testCase{
```
limhawjia, line 434: Let's add one more test case where `limitedCapacity` is not 0

pkg/controllers/scheduler/util_test.go
```diff
@@ -112,3 +120,283 @@ func TestMatchedPolicyKey(t *testing.T) {
...
+			Generation: 1,
+		},
+		Spec: fedcorev1a1.GenericFederatedObjectSpec{
+			Template: apiextensionsv1.JSON{Raw: rawTargetTemplate},
+		},
+	}
+}
+
+func Test_validatemigrationConfig(t *testing.T) {
+	sourceObject := &unstructured.Unstructured{
```
limhawjia, line 137: NIT: would prefer these test specs to be organized as a table with a suitable name for each.

pkg/controllers/statusaggregator/controller.go
```diff
@@ -347,19 +343,60 @@ func (a *StatusAggregator) reconcile(ctx context.Context, key reconcileKey) (sta
...
+			return worker.StatusError
+		}
+		sourceObject = newObj
+		needUpdate = needUpdate || newNeedUpdate
+	} else {
+		logger.V(3).Info("Plugin not found")
+	}
+
+	if needUpdate {
+		if err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
limhawjia, line 382: Why do we choose to retry on conflict here? A change in the source object might cause the status aggregated in this loop to become outdated

pkg/controllers/scheduler/planner/planner_test.go
Poor12: Added.

pkg/controllers/scheduler/util_test.go
Poor12: Updated.

pkg/controllers/statusaggregator/controller.go
Poor12: The original concern here is that the Update action has just been executed before, and I am afraid that the cache will not be refreshed in time and cause conflicts. After careful inspection, the cache is not used here. Updated.

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,39 @@ func (s *Scheduler) prepareToSchedule(
...
+	_, customMigrationInfoExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
+	if !migrationConfigExists {
+		if customMigrationInfoExists {
+			delete(fedObject.GetAnnotations(), common.AppliedMigrationConfigurationAnnotation)
+		}
+	}
+	if migrationConfigExists {
+		customMigrationInfoBytes, err := getCustomMigrationInfoBytes(existingMigrationConfigBytes, fedObject, ftc)
+		if err == nil {
+			fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation] = customMigrationInfoBytes
```
limhawjia, line 376: Apart from display purposes, what else is `AppliedMigrationConfigurationAnnotation` used for? Should it also be written if the `migrationconfig` and `rsp` plugins are disabled?

REestrella: `AppliedMigrationConfigurationAnnotation` will be used by the scheduler and evict instances through the rsp plugin. For the second question, my answer is yes. This is similiar to auto-migration. Auto-migration will also add an annotation on fedObj which indicates the estimatedCapacity whatever rsp plugin is enabled and disabled. When rsp plugin is disabled, migration does not take effect but annotation will normally be added.

Poor12: AppliedMigrationConfigurationAnnotation will be used by the scheduler and evict instances through the rsp plugin. For the second question, my answer is yes. This is similiar to auto-migration. Auto-migration will also add an annotation on fedObj which indicates the estimatedCapacity whatever rsp plugin is enabled and disabled. When rsp plugin is disabled, migration does not take effect but annotation will normally be added.

limhawjia: Why is `MigrationConfigurationAnnotation` insufficient on its own?

Poor12: One main reason is to show the user the currently applied migration config. For example, user may enter a wrong config and scheduler still use the latest migration config. Now user cannot find the applied migration config Intuitively(maybe can be found in events or logs, but it's not convenient). Another one is for user input, we would prefer it to be structured and clearer, while for internal use, the main consideration is the convenience of the program.

limhawjia (approved):
LGTM!

pkg/controllers/common/constants.go
```diff
@@ -116,6 +116,11 @@ const (
 	TemplateGeneratorMergePatchAnnotation = FederateControllerPrefix + "template-generator-merge-patch"
 
 	LatestReplicasetDigestsAnnotation = DefaultPrefix + "latest-replicaset-digests"
+
+	// MigrationConfigurationAnnotation contains custom migration configuration from users.
+	MigrationConfigurationAnnotation = DefaultPrefix + "migration-configuration"
+	// AppliedMigrationConfigurationAnnotation contains applied custom migration configuration from users.
```
gary-lgy, line 122:
```suggestion
	// AppliedMigrationConfigurationAnnotation contains the applied custom migration configuration.
```

pkg/controllers/scheduler/constants.go
```diff
@@ -45,6 +45,7 @@ const (
 	EventReasonInvalidFollowsObject      = "InvalidFollowsObject"
 	EventReasonWebhookConfigurationError = "WebhookConfigurationError"
 	EventReasonWebhookRegistered         = "WebhookRegistered"
+	EventReasonParseMigrationStrategy    = "ParseMigrationStrategy"
```
gary-lgy, line 48: We use `config` for the annotation and `strategy` here. Let's unify them to avoid confusing the users.

Poor12: Got.

pkg/controllers/scheduler/framework/plugins/rsp/rsp.go
```diff
@@ -161,6 +166,7 @@ func (pl *ClusterCapacityWeight) ReplicaScheduling(
 		ExtractClusterNames(clusters),
 		currentReplicas,
 		estimatedCapacity,
+		limitedCapacity,
```
gary-lgy, line 169: Can we combine `estimatedCapacity` and `limitedCapacity` by taking the minimum for each element before calling `planner.Plan`? The planner is already very complicated and let's not add to its complexity.

Poor12: I am afraid we cannot. Because there are a little differences. For example, limitedCapacity does not have overflow logic.

pkg/controllers/scheduler/framework/plugins/rsp/rsp_test.go
```diff
@@ -442,6 +442,13 @@ func TestClusterWeights(t *testing.T) {
 			schedulingUnit: framework.SchedulingUnit{
 				DesiredReplicas: pointer.Int64(10),
 				SchedulingMode:  fedcorev1a1.SchedulingModeDivide,
+				CustomMigration: framework.CustomMigrationSpec{
```
gary-lgy, line 445: Why add `CustomMigration` here if it doesn't have any effect?

Poor12: This is a mistake.

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,39 @@ func (s *Scheduler) prepareToSchedule(
 ) {
 	logger := klog.FromContext(ctx)
 
+	// check migration config
+
+	existingMigrationConfigBytes, migrationConfigExists := fedObject.GetAnnotations()[common.MigrationConfigurationAnnotation]
+	_, customMigrationInfoExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
+	if !migrationConfigExists {
+		if customMigrationInfoExists {
+			delete(fedObject.GetAnnotations(), common.AppliedMigrationConfigurationAnnotation)
```
gary-lgy, line 370: This relies on the fact that `fedObject.GetAnnotations()` returns a shallow copy, which is not necessarily the case for all types. e.g. if `fedObject` is an unstructured, `GetAnnotations()` returns a deep copy. A safer way to update the annotation requires calling `SetAnnotations()`.

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,39 @@ func (s *Scheduler) prepareToSchedule(
...
+		}
+	}
+	if migrationConfigExists {
+		customMigrationInfoBytes, err := getCustomMigrationInfoBytes(existingMigrationConfigBytes, fedObject, ftc)
+		if err == nil {
```
gary-lgy, line 375: It is idiomatic in golang to handle the `err != nil` case first. Is there any particular reason the `err == nil` case must be checked first?

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,39 @@ func (s *Scheduler) prepareToSchedule(
...
+			fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation] = customMigrationInfoBytes
+			logger.V(3).Info("Succeed to parse custom migration configuration")
```
gary-lgy, line 377:
```suggestion
			logger.V(3).Info("Successfully parsed custom migration configuration")
```

pkg/controllers/scheduler/scheduler.go
Poor12: I just discovered this and used the current writing method. SetAnnotation may need to add an additional temporary variable.

pkg/controllers/scheduler/scheduler.go
Poor12: No particular reason. I'd like to change the order.

pkg/controllers/scheduler/util.go
```diff
@@ -98,3 +100,93 @@ func getResourceRequest(
 	resource := resourceutil.GetPodResourceRequests(podSpec)
 	return *framework.NewResource(resource), nil
 }
+
+func getCustomMigrationInfoBytes(existingMigrationConfigBytes string,
```
gary-lgy, line 104:
```suggestion
func getCustomMigrationInfoBytes(
	existingMigrationConfigBytes string,
```

pkg/controllers/scheduler/util.go
```diff
@@ -98,3 +100,93 @@ func getResourceRequest(
...
+			return fmt.Errorf("clusterName cannot be empty")
+		}
+		if replicasMigration.LimitedCapacity == nil {
+			return fmt.Errorf("limitedCapacity cannot be empty")
+		}
+		if *replicasMigration.LimitedCapacity < 0 {
+			return fmt.Errorf("limitedCapacity cannot be less than zero")
+		}
+
+		sourceObj, err := fedObj.GetSpec().GetTemplateAsUnstructured()
```
gary-lgy, line 143: This should be done outside the for loop?

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,39 @@ func (s *Scheduler) prepareToSchedule(
 ) {
 	logger := klog.FromContext(ctx)
 
+	// check migration config
+
+	existingMigrationConfigBytes, migrationConfigExists := fedObject.GetAnnotations()[common.MigrationConfigurationAnnotation]
```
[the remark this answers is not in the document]
Poor12: We need to place it here because when applied migration config does not change we do not trigger re-schedule. Therefore parsing should be before the logic to check whether to trigger scheduling.

pkg/controllers/scheduler/util.go
Poor12: Agreed.

pkg/controllers/statusaggregator/controller.go
```diff
@@ -347,15 +342,57 @@ func (a *StatusAggregator) reconcile(ctx context.Context, key reconcileKey) (sta
...
+	plugin := plugins.GetPlugin(key.gvk)
+	if plugin != nil {
+		newObj, newNeedUpdate, err := plugin.AggregateStatuses(ctx, sourceObject, fedObject, clusterObjs, clusterObjsUpToDate)
+		if err != nil {
+			return worker.StatusError
+		}
+		sourceObject = newObj
+		needUpdate = needUpdate || newNeedUpdate
+	} else {
+		logger.V(3).Info("Plugin not found")
```
gary-lgy, line 377: This line would get spammed for resources without a type-specific plugin. Can we remove it?

pkg/controllers/statusaggregator/controller.go
```diff
@@ -347,15 +342,57 @@ func (a *StatusAggregator) reconcile(ctx context.Context, key reconcileKey) (sta
...
+	}
+
+	if needUpdate {
+		latestSourceObj, err := a.dynamicClient.Resource(ftc.GetSourceTypeGVR()).Namespace(key.namespace).
```
gary-lgy, line 381: Why do we need to get a fresh copy of the source object?

pkg/controllers/statusaggregator/plugins/common/migrationconfig.go
```diff
@@ -0,0 +1,72 @@
...
+
+	fedObjMigrationConfig, fedObjExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
+	if sourceObject.GetAnnotations() == nil {
+		if fedObjExists {
+			newAnnotation := make(map[string]string, 1)
+			newAnnotation[common.AppliedMigrationConfigurationAnnotation] = fedObjMigrationConfig
+			sourceObject.SetAnnotations(newAnnotation)
+			needUpdate = true
+		}
+		return sourceObject, needUpdate, nil
```
gary-lgy, line 52: Consider using early return for better readability.

pkg/controllers/statusaggregator/plugins/common/migrationconfig.go
```diff
@@ -0,0 +1,72 @@
...
+func (receiver *MigrationConfigPlugin) AggregateStatuses(
+	ctx context.Context,
+	sourceObject *unstructured.Unstructured,
+	fedObject fedcorev1a1.GenericFederatedObject,
+	clusterObjs map[string]interface{},
+	clusterObjsUpToDate bool,
+) (*unstructured.Unstructured, bool, error) {
+	needUpdate := false
+
+	fedObjMigrationConfig, fedObjExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
```
gary-lgy, line 44:
```suggestion
	fedObjMigrationConfig, fedAnnotationExists := fedObject.GetAnnotations()[common.AppliedMigrationConfigurationAnnotation]
```

pkg/controllers/statusaggregator/controller.go
Poor12: Because we just update the sourceObj early. So we need to get the latest obj before updateStatus.

pkg/controllers/statusaggregator/controller.go
Poor12: Agreed.

pkg/controllers/scheduler/scheduler.go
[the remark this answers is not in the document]
gary-lgy: Understood. This looks like a gotcha. Can we add a comment to explain the above for future developers so they do not accidentally move it to `applySchedulingResults`?

pkg/controllers/statusaggregator/controller.go
gary-lgy: Understood. Would using `patch` avoid the additional get? I'm slightly worried about the burden on the apiserver.

Poor12: My concern is that users may update the annotation on source object in the same time when statusAggregator updates the anno. Using patch may override the changes users made.

pkg/controllers/scheduler/scheduler.go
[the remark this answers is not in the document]
Poor12: done.

pkg/controllers/scheduler/util.go
gary-lgy: nit: we don't need to unmarshal the template if `len(migrationConfig.ReplicasMigration) == 0`

pkg/controllers/scheduler/scheduler.go
```diff
@@ -361,6 +361,38 @@ func (s *Scheduler) prepareToSchedule(
 ) {
 	logger := klog.FromContext(ctx)
 
+	// check migration config
+	// We must check migration configuration before the judgement of whether to skip scheduling.
+	// Because whether AppliedMigrationConfigurationAnnotation changes is one of the factors
+	// that affects whether rescheduling is needed.
```
gary-lgy, line 367:
```suggestion
	// update applied migration config
	// We must update migration configuration before the checking scheduling triggers because
	// the former is part of the latter.
```

PR #267, opened 2023-11-10
fix(scheduler): add terminating filter

JackZxj:
For the clusters who are in terminating, we can just filter them out directly for the following reasons:
* The sync controller will not dispatch or delete resources from terminating clusters, unless cascading deletion is enabled.
* Considering that the user has submitted the request to delete member clusters, it is expected and more stable to directly filter out when scheduling is triggered.

### DIFF
--- a/pkg/controllers/scheduler/framework/plugins/clusterterminating/clusterterminating.go
+++ b/pkg/controllers/scheduler/framework/plugins/clusterterminating/clusterterminating.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2023 The KubeAdmiral Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


pkg/controllers/scheduler/framework/plugins/clusterterminating/clusterterminating.go
```diff
@@ -0,0 +1,53 @@
...
+	cluster *fedcorev1a1.FederatedCluster,
+) *framework.Result {
+	err := framework.PreCheck(ctx, su, cluster)
+	if err != nil {
+		return framework.NewResult(framework.Error, err.Error())
+	}
+
+	// Prevent scheduling to termianing cluster.
+	if !cluster.DeletionTimestamp.IsZero() {
+		return framework.NewResult(framework.Unschedulable, "cluster is terminating")
```
Poor12, line 49:
```suggestion
		return framework.NewResult(framework.Unschedulable, "cluster(s) were terminating")
```
Optimize the prompts given to users.

mrlihanbo (approved):

limhawjia (approved):

PR #270, opened 2023-11-17
refactor: replace runtime.Object with metav1.Object in pkg/util/finalizers

### LINKED ISSUE #99: `pkg/controllers/util/annotation/annotation.go` should use `metav1.Object` instead of `pkgruntime.Object` for more ergonomic annotation handling

z1cheng:
This PR is to replace runtime.Object with metav1.Object in `pkg/util/finalizers`

related to #99

### DIFF
--- a/pkg/controllers/sync/controller.go
+++ b/pkg/controllers/sync/controller.go
@@ -624,12 +624,7 @@ func (s *SyncController) syncToClusters(
 			)
 			continue
 		}
-		hasFinalizer, err := finalizersutil.HasFinalizer(cluster, FinalizerCascadingDelete)
-		if err != nil {
-			shouldRecheckAfterDispatch = true
-			dispatcher.RecordClusterError(fedcorev1a1.FinalizerCheckFailed, clusterName, err)
-			continue
-		}
+		hasFinalizer := finalizersutil.HasFinalizer(cluster, FinalizerCascadingDelete)
 		if !hasFinalizer {
 			// we should not sync before finalizer is added
 			shouldRecheckAfterDispatch = true
@@ -971,9 +966,9 @@ func (s *SyncController) ensureFinalizer(ctx context.Context, fedResource Federa
 	ctx, keyedLogger := logging.InjectLoggerValues(ctx, "finalizer-name", FinalizerSyncController)
 
 	obj := fedResource.Object()
-	isUpdated, err := finalizersutil.AddFinalizers(obj, sets.NewString(FinalizerSyncController))
-	if err != nil || !isUpdated {
-		return err
+	isUpdated := finalizersutil.AddFinalizers(obj, sets.New(FinalizerSyncController))
+	if !isUpdated {
+		return nil
 	}
 
 	keyedLogger.V(1).Info("Adding finalizer to federated object")
@@ -995,9 +990,9 @@ func (s *SyncController) removeFinalizer(ctx context.Context, fedResource Federa
 	ctx, keyedLogger := logging.InjectLoggerValues(ctx, "finalizer-name", FinalizerSyncController)
 
 	obj := fedResource.Object()
-	isUpdated, err := finalizersutil.RemoveFinalizers(obj, sets.NewString(FinalizerSyncController))
-	if err != nil || !isUpdated {
-		return err
+	isUpdated := finalizersutil.RemoveFinalizers(obj, sets.New(FinalizerSyncController))
+	if !isUpdated {
+		return nil
 	}
 
 	keyedLogger.V(1).Info("Removing finalizer from federated object")
@@ -1024,9 +1019,9 @@ func (s *SyncController) ensureClusterFinalizer(ctx context.Context, cluster *fe
 		if err != nil {
 			return err
 		}
-		isUpdated, err := finalizersutil.AddFinalizers(cluster, sets.NewString(FinalizerCascadingDelete))
-		if err != nil || !isUpdated {
-			return err
+		isUpdated := finalizersutil.AddFinalizers(cluster, sets.New(FinalizerCascadingDelete))
+		if !isUpdated {
+			return nil
 		}
 		cluster, err = s.fedClient.CoreV1alpha1().FederatedClusters().Update(ctx, cluster, metav1.UpdateOptions{})
 		return err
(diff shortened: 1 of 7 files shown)


pkg/util/finalizers/finalizers.go
```diff
@@ -22,49 +22,36 @@ are Copyright 2023 The KubeAdmiral Authors.
...
 
 // RemoveFinalizers removes the given finalizers from the given objects ObjectMeta.
 // Returns true if the object was updated.
-func RemoveFinalizers(obj runtime.Object, finalizers sets.String) (bool, error) {
-	accessor, err := meta.Accessor(obj)
-	if err != nil {
-		return false, err
-	}
-	oldFinalizers := sets.NewString(accessor.GetFinalizers()...)
+func RemoveFinalizers(obj metav1.Object, finalizers sets.String) (bool, error) {
```
z1cheng, line 49: Wondering if the returned error is still required, since it's always `nil`. Please let me know if it should be removed :)

SOF3: applies to `AddFinalizers` too

z1cheng: @SOF3 Updated, please have a look when you are available

pkg/controllers/sync/controller.go
```diff
@@ -971,9 +966,9 @@ func (s *SyncController) ensureFinalizer(ctx context.Context, fedResource Federa
 	ctx, keyedLogger := logging.InjectLoggerValues(ctx, "finalizer-name", FinalizerSyncController)
 
 	obj := fedResource.Object()
-	isUpdated, err := finalizersutil.AddFinalizers(obj, sets.NewString(FinalizerSyncController))
-	if err != nil || !isUpdated {
-		return err
+	isUpdated := finalizersutil.AddFinalizers(obj, sets.NewString(FinalizerSyncController))
+	if !isUpdated {
+		return fmt.Errorf("failed to add finalizer to federated object")
```
gary-lgy, line 971: This looks wrong. Pls check again.

pkg/controllers/sync/controller.go
```diff
@@ -995,9 +990,9 @@ func (s *SyncController) removeFinalizer(ctx context.Context, fedResource Federa
 	ctx, keyedLogger := logging.InjectLoggerValues(ctx, "finalizer-name", FinalizerSyncController)
 
 	obj := fedResource.Object()
-	isUpdated, err := finalizersutil.RemoveFinalizers(obj, sets.NewString(FinalizerSyncController))
-	if err != nil || !isUpdated {
-		return err
+	isUpdated := finalizersutil.RemoveFinalizers(obj, sets.NewString(FinalizerSyncController))
+	if !isUpdated {
+		return fmt.Errorf("failed to remove finalizer from federated object")
```
gary-lgy, line 995: Same as above.

pkg/controllers/sync/controller.go
```diff
@@ -1024,9 +1019,9 @@ func (s *SyncController) ensureClusterFinalizer(ctx context.Context, cluster *fe
 		if err != nil {
 			return err
 		}
-		isUpdated, err := finalizersutil.AddFinalizers(cluster, sets.NewString(FinalizerCascadingDelete))
-		if err != nil || !isUpdated {
-			return err
+		isUpdated := finalizersutil.AddFinalizers(cluster, sets.NewString(FinalizerCascadingDelete))
+		if !isUpdated {
+			return fmt.Errorf("failed to add finalizer to federated cluster")
```
gary-lgy, line 1024: Same as above.

pkg/controllers/sync/resource.go
```diff
@@ -198,9 +198,7 @@ func (r *federatedResource) ObjectForCluster(clusterName string) (*unstructured.
 }
 
 func addRetainObjectFinalizer(obj *unstructured.Unstructured) error {
-	if _, err := finalizers.AddFinalizers(obj, sets.NewString(dispatch.RetainTerminatingObjectFinalizer)); err != nil {
-		return err
-	}
+	finalizers.AddFinalizers(obj, sets.NewString(dispatch.RetainTerminatingObjectFinalizer))
 	return nil
```
gary-lgy, line 202: Looks like we can remove this redundant nil return.

gary-lgy:
@z1cheng Could you amend the PR description to remove the `close` in `close #99`? This PR doesn't fully close it, since the annotation util functions have not been fully refactored yet.

pkg/controllers/sync/controller.go
z1cheng: The original logic is that if `err` is nil and `isUpdated` is false, the err (nil) will be returned here. So let me update it to match the original logic, thanks!

pkg/controllers/sync/dispatch/unmanaged.go
```diff
@@ -132,17 +132,8 @@ func (d *unmanagedDispatcherImpl) Delete(
...
-				wrappedErr := d.wrapOperationError(err, clusterName, op)
-				keyedLogger.Error(wrappedErr, "Failed to delete target object in cluster")
-			} else {
-				d.recorder.recordOperationError(ctx, fedcorev1a1.DeletionFailed, clusterName, op, err)
-			}
-			result = false
-			return result
-		}
+		needUpdate := removeRetainObjectFinalizer(clusterObj)
+		var err error
```
SOF3, line 136: Why do we declare `err` here? The next immediate use is only inside the `needUpdate` block.

z1cheng: image
@SOF3 if not declaring `err` here,  we need to considering removing/renaming the unused variable `clusterObj`, but it looks like `clusterObj` wants to be overwritten by `clusterObj, err = client.Resource(d.targetGVR).Namespace(targetName.Namespace).Update(xxx)` in the original logic

SOF3: @z1cheng I mean you can move it one line downwards, since this `err` is only used inside the block, and change line 171 to `err :=`

z1cheng: Updated

z1cheng:
Submit a commit to replace the deprecated `sets.String` with generic `sets.Set`

pkg/util/finalizers/finalizers.go
```diff
@@ -28,30 +28,31 @@ import (
...
-func RemoveFinalizers(obj metav1.Object, finalizers sets.String) bool {
-	oldFinalizers := sets.NewString(obj.GetFinalizers()...)
+func RemoveFinalizers(obj metav1.Object, finalizers sets.Set[string]) bool {
+	oldFinalizers := sets.New(obj.GetFinalizers()...)
 	if oldFinalizers.Intersection(finalizers).Len() == 0 {
 		return false
 	}
 	newFinalizers := oldFinalizers.Difference(finalizers)
-	obj.SetFinalizers(newFinalizers.List())
+	// obj.SetFinalizers(newFinalizers.UnsortedList())
```
gary-lgy, line 55: Pls remove this commented out line

gary-lgy:
@z1cheng pls rebase the PR (and keep it rebased), address the comment and then LGTM. Don't worry about the codecov/patch check failure.

pkg/util/finalizers/finalizers.go
z1cheng: Updated, thanks!

SOF3 (approved):

gary-lgy (approved):

z1cheng:
@gary-lgy @SOF3 This looks like it can be merged, please let me know what else is needed :)

PR #287, opened 2023-11-23
doc: update quick start

JackZxj:


docs/quickstart.md
```diff
@@ -172,95 +181,116 @@ If the KubeAdmiral control plane is working properly, the deployment will be pro
...
-echo-server   6/6     6            6           10m
+echo-server   6/6     6            6           26s
 ```
 
-We can also view the status of the deployments in each member cluster by checking its `FederatedDeploymentStatus` object.
+We can also view the status of the deployments in each member cluster by checking its `CollectedStatus` object.
 
 ```console
-$ kubectl get fdeploystatus echo-server -oyaml
+# Usually the name consists of the resource name and its ResourcesGroup,
```
gary-lgy, line 190:
```suggestion
# Usually the name consists of the resource name and its API group (e.g. echo-server-deployments.apps), 
```

docs/quickstart.md
```diff
@@ -172,95 +181,116 @@ If the KubeAdmiral control plane is working properly, the deployment will be pro
...
+# but sometimes there may be a hash suffix to avoid duplication.
```
gary-lgy, line 191:
```suggestion
# but sometimes there may be a hash suffix to avoid duplication.
# We can use `get collectedstatus -l apps/v1=Deployment` to get all `CollectedStatus`es for `apps/v1.Deployment`s.
```

docs/quickstart.md
SOF3: This is a `console` block, where lines starting with `#` and `$` are treated as command lines (for root and non-root users respectively). `#` does not represent comments here.

gary-lgy (approved):

PR #290, opened 2023-12-01
refactor: remove redundant returned error

### LINKED ISSUE #99: `pkg/controllers/util/annotation/annotation.go` should use `metav1.Object` instead of `pkgruntime.Object` for more ergonomic annotation handling

z1cheng:
close #99

### DIFF
--- a/pkg/util/annotation/annotation.go
+++ b/pkg/util/annotation/annotation.go
@@ -17,9 +17,6 @@ limitations under the License.
 package annotation
 
 import (
-	"fmt"
-	"reflect"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/util/annotation/annotation.go
```diff
@@ -30,81 +29,71 @@ const (
 )
 
 // HasAnnotationKey returns true if the given object has the given annotation key in its ObjectMeta.
-func HasAnnotationKey(obj metav1.Object, key string) (bool, error) {
+func HasAnnotationKey(obj metav1.Object, key string) bool {
 	if IsNilPointer(obj) {
```
gary-lgy, line 33: Let's remove this `IsNilPointer` check to expose the problem early.

pkg/util/annotation/annotation.go
```diff
@@ -30,81 +29,71 @@ const (
...
 // Returns true if the object was updated.
-func AddAnnotation(obj metav1.Object, key, value string) (bool, error) {
-	if IsNilPointer(obj) {
-		return false, fmt.Errorf("object(%T) is nil pointer", obj)
-	}
-
-	if key == "" {
-		return false, fmt.Errorf("key is a empty string.")
+func AddAnnotation(obj metav1.Object, key, value string) bool {
+	if IsNilPointer(obj) || key == "" {
```
gary-lgy, line 61: The key is usually a constant, so the chance that it's empty is slim. Let's remove this `key` check to make it clear that it's the caller's responsibility to pass in a valid key.

pkg/util/annotation/annotation.go
z1cheng: Removed

pkg/util/annotation/annotation.go
z1cheng: Removed

pkg/util/annotation/annotation.go
gary-lgy: Let's try removing `IsNilPointer` check from all the functions in this file. If there's no other references to `IsNilReference`, we should remove its definition as well.

pkg/util/annotation/annotation.go
gary-lgy: Pls fix the unit tests.

PR #291, opened 2023-12-01
feat: add prioritySort plugin

Poor12:


mrlihanbo (approved):

pkg/controllers/scheduler/framework/runtime/framework.go
```diff
@@ -246,9 +247,13 @@ func (f *frameworkImpl) RunSelectClustersPlugin(
 				result.Message(),
 			)
 			klog.Error(msg)
-			return clusters, framework.NewResult(framework.Error, msg)
+			return nil, framework.NewResult(framework.Error, msg)
 		}
-		return clusters, result
+		clusterScores = newClusterScores
```
gary-lgy, line 252: If `MaxClusters` doesn't sort internally, the framework needs to do the sorting. Otherwise in the absence of `PrioriotySort`, `MaxClusters` will stop working correctly.

pkg/controllers/scheduler/framework/plugins/prioritysort/priority_sort.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (pl *PrioritySort) SelectClusters(
+	ctx context.Context,
+	su *framework.SchedulingUnit,
+	clusterScoreList framework.ClusterScoreList,
+) (framework.ClusterScoreList, *framework.Result) {
+	var ret framework.ClusterScoreList
+
+	if len(su.Priorities) > 0 || (su.SchedulingMode == fedcorev1a1.SchedulingModeDivide &&
```
gary-lgy, line 45: Why is the second part of this check necessary?

Poor12: For preferenceBinpack, using score as priority is not stable. For example, score is now mainly from resource usage and utilization. But through the binpack process, priority of different clusters may change which may bring some unexpected results

limhawjia:
This PR changes the scheduler's framework to support having multiple select plugins. However, it changes the select plugin interface such that it can now override the cluster scores list. I have some reservations about this.

1. There is now an overlap of responsibilities with the scoring stage.
2. It gives a single select plugin too much power. The scores returned by a select plugin will completely replace any existing ones, possibly rendering the work done by previous plugins or even the whole scoring stage obsolete.

pkg/controllers/scheduler/framework/runtime/framework.go
gary-lgy: A plugin shouldn't depend on another plugin to function.

pkg/controllers/scheduler/framework/plugins/prioritysort/priority_sort.go
gary-lgy: Pls point this out in the comment.

pkg/controllers/scheduler/framework/runtime/framework.go
Poor12: I got your points. There will still be some depency if we want multiple select plugins. For example, priority must be before maxClusters. Let's think about refacter the framework later.

PR #289, opened 2023-11-29
feat: add binpack api

Poor12:

PR #295, opened 2023-12-08
add preference binpack replicas plugin

mrlihanbo:


pkg/apis/core/v1alpha1/extensions_schedulingprofile.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/kubewharf/kubeadmiral/pkg/controllers/scheduler/framework/plugins/names"
 )
 
-func GetDefaultEnabledPlugins() *fedcore.EnabledPlugins {
+func GetDefaultEnabledPlugins(replicasPlugin string) *fedcore.EnabledPlugins {
```
gary-lgy, line 24: Any chance to mutate the returned `EnabledPlugins` in the scheduler? The default list is the baseline, and should not vary depending on configuration.

pkg/controllers/scheduler/framework/plugins/preferencebinpack/planner.go
```diff
@@ -0,0 +1,331 @@
...
+			currentTotalScheduledReplicas += scheduled
+		}
+	}
+
+	if !keepUnschedulableReplicas && currentTotalScheduledReplicas == totalReplicas {
+		klog.V(4).Infof("Trim overflow replicas when getting desired number of scheduled replicas")
+		return scheduledReplicas, nil, nil
+	}
+
+	// If we don't desiredPlan to avoid migration, just return the plan computed from preferences
```
gary-lgy, line 121:
```suggestion
	// If we don't need to avoid migration, just return the plan computed from preferences
```

pkg/controllers/scheduler/framework/plugins/preferencebinpack/planner.go
```diff
@@ -0,0 +1,331 @@
...
+
+	var desiredTotalReplicas int64
+	for _, replicas := range desiredPlan {
+		desiredTotalReplicas += replicas
+	}
+
+	if !keepUnschedulableReplicas {
+		newOverflow := make(map[string]int64)
+		for key, value := range desiredOverflow {
+			value = framework.MinInt64(value, totalReplicas-currentTotalScheduledReplicas)
```
gary-lgy, line 134: Let's add a comment - 
```golang
// Cap the overflow at currently unscheduled replicas.
```

pkg/controllers/scheduler/framework/plugins/preferencebinpack/planner.go
```diff
@@ -0,0 +1,331 @@
...
+	scaleUpCount int64,
+	availableClusters []string,
+) (map[string]int64, error) {
+	namedPreferences := make([]*namedClusterPreferences, 0, len(availableClusters))
+	for _, cluster := range availableClusters {
+		current := currentReplicaCount[cluster]
+		desired := desiredReplicaCount[cluster]
+		if desired > current {
+			pref := &namedClusterPreferences{
+				clusterName: cluster,
```
gary-lgy, line 239: Should the weights in the new `pref` equal those in `rsp`? When scaling up, we still want to fill clusters with higher priority.

pkg/controllers/scheduler/framework/plugins/preferencebinpack/planner.go
```diff
@@ -0,0 +1,331 @@
...
+	// no estimatedCapacity and hence no overflow
+	replicasToScaleDown, _ := getDesiredPlan(namedPreferences, nil, nil, scaleDownCount)
+	for cluster, count := range replicasToScaleDown {
+		currentReplicaCount[cluster] -= count
+	}
+
+	return currentReplicaCount, nil
+}
+
+func preCheck(
```
gary-lgy, line 288:
```suggestion
func shouldReplan(
```

pkg/controllers/scheduler/framework/plugins/preferencebinpack/planner.go
```diff
@@ -0,0 +1,331 @@
...
+	currentReplicaCount, estimatedCapacity, scheduledReplicas map[string]int64,
+) (available bool) {
+	defer func() {
+		if !available {
+			klog.V(4).Infof("Current estimate capacity is unavailable")
+		}
+	}()
+
+	for cluster, replicas := range scheduledReplicas {
+		if capacity, exist := estimatedCapacity[cluster]; exist && capacity != replicas {
```
gary-lgy, line 323: Let's add a comment here to explain why we skip planning here
```golang
If `capacity != replicas`, some replicas are still being scheduled. We refer planning until the status of all replicas are known.
```

pkg/controllers/automigration/util.go
```diff
@@ -27,18 +27,21 @@ import (
...
-) (unschedulableCount int, nextCrossIn *time.Duration) {
+) (scheduledCount, unschedulableCount int, nextCrossIn *time.Duration) {
 	for _, pod := range podList {
 		if pod.GetDeletionTimestamp() != nil {
 			continue
 		}
 
 		scheduledCondition, isUnschedulable := getPodScheduledCondition(pod)
 		if !isUnschedulable {
+			if scheduledCondition != nil {
```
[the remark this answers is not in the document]
mrlihanbo: lgtm, I will add the additionally check

limhawjia:
Adding on to @gary-lgy 's point, the scheduling profile used to dictate which fields in `PropagationPolicy` are respected and what semantics they had. In a sense, the scheduling profile governed `PropagationPolicy`.  There used to be a clean one-way relation between profiles and policies.

This PR introduces the inverse of this, creating a two-way relation between profiles and policies where they can affect one another in elaborate and unexpected ways. This is bad design and a slippery slope in my opinion as well.

limhawjia:
I propose an alternative solution. @mrlihanbo @gary-lgy  

Instead of introducing a new `ReplicasStrategy` field that modifies the plugins used in the scheduling profile, how about we provide multiple built-in profiles? One of which is `spread` (the default), and one of which is `binpack`.

`spread` will consist of the current default list of plugins, while `binpack` will contain plugins configured to provide bin-packing capabilities. Doing so will retain ease-of-use for the user. They do not have to configure their own scheduling profile, and enabling bin-packing is still a matter of setting a single field (just `schedulingProfile` instead of `replicasStrategy`).

Poor12:
Agreed all. But there is one thing, schedulingProfile is indeed more complicated than adding replicaStrategy to policy. I think if preferenceBinpack is a basic feature, it would be better to add replicaStrategy. Otherwise, if it's a extended feature, it would be better to add it in schedulingProfile. During the former discussion, seems we hope it to be a basic feature?

gary-lgy:
To those without context, there was an offline discussion among @manmanhappy @gary-lgy @mrlihanbo @JackZxj and @Poor12 on this topic. @gary-lgy proposed the 2 built-in profiles solution mentioned by @limhawjia above. However, @manmanhappy rejected it and proposed the current solution as implemented in #289 and this PR (and did not leave much room for further discussion).

Replying to @Poor12's comment:
> Agreed all. But there is one thing, schedulingProfile is indeed more complicated than adding replicaStrategy to policy.
From whose perspective? The profiles are built-in (either in code or created by the platform). From the users' perspective, they need to configure a single field in the policy spec, be it `pp.spec.schedulingProfile = binpack` or `pp.spec.replicasStrategy = binpack`. I do not see how one is "more complicated" than the other. To the developers, profile is an existing mechanism and the current solution involves API changes, so the latter is more complicated.
Most importantly, the current solution pollutes the scheduler architecture for reasons stated in previous comments. 

> I think if preferenceBinpack is a basic feature, it would be better to add replicaStrategy. Otherwise, if it's as extended feature, it would be better to add it in schedulingProfile.
Basic or extended is subjective, and I don't think it's the right yardstick to determine whether a configuration option should become a policy field. There are in fact many considerations for preferring the profile approach:
1. We already have too many fields in the policy spec. We can't expect the users to keep all of them in their head. Adding more fields further worsens the cognitive load of using a policy. Certainly, we can use defaulting to reduce some of its complexity, but managing the fields for us is also a burden as the interactions between them become ever harder to reason with. 
2. This new replica scheduling behaviour only works well for serverless/on-demand Kubernetes clusters and will spell disaster if used on a normal cluster backed by reversed nodes. How many percent of our user space would need to use this niche feature? Does it justify adding a new policy field when we already have another solution (profile) which can satisfy the requirements?

To be absolutely clear, I maintain my position (which I made clear during the initial offline discussion) that we shouldn't go down this slippery slope of continuously adding new fields.

Poor12 (approved):

PR #304, opened 2024-01-08
feat: support mcs

Poor12:


hack/generate-groups.sh
```diff
@@ -67,7 +67,9 @@ ${GOBIN}/controller-gen crd paths=$(codegen::join ";" "${INPUT_BASE}/${groups[0]
 
 # patch CRDs with no-federate annotation
 for crd_file in config/crds/*.yaml; do
-  yq eval -i ".metadata.annotations[\"${NO_FEDERATED_ANNOTATION}\"] = \"true\"" "${crd_file}"
+  if [[ $crd_file =~ ^config/crds/core.kubeadmiral.io.* ]]; then
```
gary-lgy, line 70: This can get messy over time. Consider using directories to organise CRDs.

pkg/controllers/common/constants.go
```diff
@@ -194,12 +211,15 @@ var (
 	DeploymentGVR = appsv1.SchemeGroupVersion.WithResource(DeploymentResource)
 	DaemonSetGVR  = appsv1.SchemeGroupVersion.WithResource(DaemonSetResource)
 	ReplicaSetGVR = appsv1.SchemeGroupVersion.WithResource(ReplicaSetResource)
+
+	ServiceExportGVR = mcsv1alpha1.SchemeGroupVersion.WithResource(ServiceExportResource)
+	EndpointSliceGVR = discoveryv1b1.SchemeGroupVersion.WithResource(EndpointSliceResource)
 )
 
-// MaxFederatedObjectNameLength defines the max length of a federated object name.
+// MaxNameLength defines the max length of a kubernetes resource name.
```
gary-lgy, line 219: This length only applies to CRDs.

Poor12: I tested some of Kubernetes native resources. Seems their max name length is also 253. I'm not sure which resources are not this length.

pkg/util/informermanager/federatedinformermanager.go
```diff
@@ -90,28 +99,33 @@ func NewFederatedInformerManager(
...
 		),
-		podListerSemaphore:    semaphore.NewWeighted(3), // TODO: make this configurable
-		initialClusters:       sets.New[string](),
-		podEventHandlers:      []*ResourceEventHandlerWithClusterFuncs{},
-		podEventRegistrations: map[string]map[*ResourceEventHandlerWithClusterFuncs]cache.ResourceEventHandlerRegistration{},
+		podListerSemaphore:              semaphore.NewWeighted(3), // TODO: make this configurable
+		initialClusters:                 sets.New[string](),
+		podEventHandlers:                []*ResourceEventHandlerWithClusterFuncs{},
+		podEventRegistrations:           map[string]map[*ResourceEventHandlerWithClusterFuncs]cache.ResourceEventHandlerRegistration{},
+		serviceExportEventHandlers:      []*FilteringResourceEventHandlerWithClusterFuncs{},
```
gary-lgy, line 125: InformerManager is shared across all components and ideally should not contain hardcoded resource-specific logic like this (the pod case above is a special case due to the volume of pods in large clusters). Any chance to abstract this further and shield InformerManager from controller-specific logic?

pkg/util/informermanager/interface.go
```diff
@@ -173,6 +226,13 @@ type FederatedInformerManager interface {
 	GetPodLister(cluster string) (lister corev1listers.PodLister, informerSynced cache.InformerSynced, exists bool)
 	GetNodeLister(cluster string) (lister corev1listers.NodeLister, informerSynced cache.InformerSynced, exists bool)
 
+	AddServiceExportEventHandler(handler *FilteringResourceEventHandlerWithClusterFuncs)
```
gary-lgy, line 229: Like stated above, expanding an interface like this is unsustainable.

pkg/util/informermanager/federatedinformermanager.go
Poor12: For dynamicInformerFactory, I did consider encapsulating a generic one. The reason for using this form is firstly because of the reference to the above-mentioned podEventHandlers, and secondly because this structure is already a bit complicated, map[string]map[*ResourceEventHandlerWithClusterFuncs]cache.ResourceEventHandlerRegistration{}, and adding another layer up will make it even more complicated. If the pod case is a special case, I think a generic logic is ok.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -0,0 +1,493 @@
...
+		return
+	}
+	logger.Info("Caches are synced")
+	c.worker.Run(ctx)
+	<-ctx.Done()
+}
+
+func (c *ServiceExportController) reconcile(ctx context.Context, reconcileKey reconcileKey) (status worker.Result) {
+	key := reconcileKey.String()
+	ctx, keyedLogger := logging.InjectLoggerValues(ctx, "control-loop", "reconcile", "object", key)
```
gary-lgy, line 224: Consider injecting the fields in `reconcileKey` individually for easier lookup.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	if !needReconcile {
+		return nil
+	}
+
+	seLister, synced, exists := c.fedInformerManager.GetServiceExportLister(key.cluster)
+	if !exists {
+		return fmt.Errorf("informer of serviceExport not exists for cluster %s", key.cluster)
+	}
+	if !cache.WaitForCacheSync(ctx.Done(), synced) {
```
gary-lgy, line 257: Will this ever timeout? More importantly, this goroutine will be occupied until this function returns. Potentially all available `ReconcileWorker` goroutines might be stuck here and new events will not be processed.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -0,0 +1,493 @@
...
+		}
+
+		importedEps := eps.DeepCopy()
+		importedEps.ObjectMeta = metav1.ObjectMeta{
+			Namespace: eps.Namespace,
+			Name:      naming.GenerateImportedEndpointSliceName(eps.GetName(), key.cluster),
+			Labels: map[string]string{
+				discoveryv1b1.LabelServiceName:            key.name,
+				EndpointSliceSourceClusterLabel:           key.cluster,
+				managedlabel.ManagedByKubeAdmiralLabelKey: managedlabel.ManagedByKubeAdmiralLabelValue,
```
gary-lgy, line 310: Having a resource with managed label in the host cluster can be very confusing. Any change to use a different label and not overload the kubeadmiral managed label?

pkg/controllers/mcs/service_import_controller.go
```diff
@@ -0,0 +1,595 @@
...
+		return worker.StatusError
+	}
+	serviceImportObj := &mcsv1alpha1.ServiceImport{}
+	if err = runtime.DefaultUnstructuredConverter.FromUnstructured(serviceImportTemplate.UnstructuredContent(), serviceImportObj); err != nil {
+		logger.Error(err, "Failed to convert ServiceImport from unstructured object")
+		return worker.StatusError
+	}
+
+	if err := c.syncPlacementsToEndpointSlice(ctx, serviceImportObj,
+		siFedObject.GetSpec().GetControllerPlacement(PrefixedGlobalSchedulerName)); err != nil {
```
gary-lgy, line 272: Should we get scheduler placements or the union of all placements?

pkg/controllers/sync/controller.go
```diff
@@ -596,6 +597,14 @@ func (s *SyncController) syncToClusters(
 			}
 		}
 
+		if clusterObj != nil {
+			// If one of fedObject and clusterObj is a derived service and the other is not,
+			// the obtained clusterObj does not correspond to fedObj and should be set to nil.
+			if util.IsDerivedService(fedResource.Object().GetAnnotations()) != util.IsDerivedService(clusterObj.GetAnnotations()) {
```
gary-lgy, line 603: I'm having a hard time understanding when this would be the case. If we reach this line, the cluster service must have the kubeadmiral managed label and must have been synced by us. Hence they must have the same `derived` status. Could you point out the flaw in my logic?

pkg/controllers/common/constants.go
gary-lgy: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

pkg/controllers/sync/controller.go
Poor12: Now there will be two fedObj, their template is the svc with the same name, the first is fedObj generated by sourceObj for foo, the second is fedObj with derived prefix generated by serviceImport for foo, we must make sure to delete the second fedObj (even if it does not take effect due to conflicts), the first type of svc issued will not be cleared. Therefore, it is necessary to provide a mechanism to mark which FedObj issued the svc of the member cluster. Considering that this situation only occurs under mcs, a special label is used.

pkg/util/informermanager/federatedinformermanager.go
Poor12: I agreed to use a common interfaces for dynamic resources. For native resources(such as endpointSlice or service), I prefer to add a separate interface. Because even if we use a common interface, we have to implement it by a switch-case logic. 
```
case endpointSliceGVK:
    factory.Discovery.V1beta1().EndpointSlice()....
case serviceGVK:
    factory.Core().V1().Services()...
```

pkg/controllers/mcs/service_import_controller.go
Poor12: Currently, serviceImport will only be scheduled by global-scheduler(follower-controller cannot). I think get scheduler placement is OK.

pkg/controllers/mcs/service_export_controller.go
Poor12: Got it. I had not discovered this hidden problem before.

config/crds/bases/core_kubeadmiral_io/core.kubeadmiral.io_clustercollectedstatuses.yaml
gary-lgy, line 1: Why `bases`? And why `core_kubeadmiral_io` instead of `core.kubeadmiral.io`?

Poor12: Bases is a standard commonly used by kubebuilder. See https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/controller-gen.md.
The folder name cannot contain ., so I changed it to _.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -0,0 +1,503 @@
...
+	"github.com/kubewharf/kubeadmiral/pkg/util/naming"
+	"github.com/kubewharf/kubeadmiral/pkg/util/worker"
+)
+
+const (
+	ServiceExportControllerName = "serviceexport-controller"
+)
+
+const (
+	EndpointSliceSourceClusterLabel = common.DefaultPrefix + "source-cluster"
```
JackZxj, line 56: cluster name might exceed 63 characters which is illegal for label value

Poor12: Updated.

pkg/util/informermanager/federatedinformermanager.go
```diff
@@ -431,6 +486,43 @@ func (m *federatedInformerManager) GetNodeLister(
...
+	gvr schema.GroupVersionResource,
+	handler *FilteringResourceEventHandlerWithClusterFuncs,
+) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.resourceEventHandlers[gvr] = append(m.resourceEventHandlers[gvr], handler)
+}
+
+func (m *federatedInformerManager) GetResourceListerFromFactory(
```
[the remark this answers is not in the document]
Poor12: For native resources, I hope to use the native informerFactory, and for dynamic resources, use dynamicInformerFactory to reduce the consumption caused by type conversion (json marshal/unmarshal). So interface{} is used here. In addition, the objects called by these interfaces usually must be preset before start. There is no chance to call the start method for other resources, so they will not be synchronized smoothly.

pkg/util/informermanager/federatedinformermanager.go
```diff
@@ -303,6 +335,25 @@ func (m *federatedInformerManager) processCluster(
...
+		for _, handler := range m.resourceEventHandlers[gvr] {
+			if gvrRegistrations[handler] == nil {
+				copied := handler.copyWithClusterName(clusterName)
+				if gvr == common.EndpointSliceGVR {
+					if r, err := factory.Discovery().V1beta1().EndpointSlices().Informer().AddEventHandler(copied); err == nil {
+						gvrRegistrations[handler] = r
+					}
+				} else {
+					if r, err := dynamicFactory.ForResource(gvr).Informer().AddEventHandler(copied); err == nil {
+						gvrRegistrations[handler] = r
```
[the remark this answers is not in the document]
Poor12: Currently only endpointSliceGVR and seviceExportGVR. And they will never be nil.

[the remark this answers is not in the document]
JackZxj: but you have exported AddResourceEventHandler, which gives it a chance to add a new gvr handler if anyone forgets

pkg/controllers/sync/controller.go
```diff
@@ -934,25 +951,33 @@ func (s *SyncController) handleDeletionInClusters(
 			continue
 		}
 		if clusterObj == nil {
+			clustersNoNeedHandleDeletion.Insert(clusterName)
+			continue
+		}
+		// If one of fedObject and clusterObj is a derived service and the other is not,
+		// the obtained clusterObj does not correspond to fedObj and should be set to nil.
```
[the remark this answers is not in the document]
Poor12: Some samples? It may be hard to give a satisfied comment. :(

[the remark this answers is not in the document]
JackZxj: This comment does not quite match the operation here, nothing to be set to nil. I'm worried that if I see this comment in a few weeks later, I might forget the reason and be confused.

JackZxj (approved):

PR #310, opened 2024-01-16
feat: add support for katalyst

JackZxj:
Two new parameters have been added to support [katalyst](https://github.com/kubewharf/katalyst-core):

* `--enable-katalyst-support=true` Enable katalyst support in the cluster controller and scheduler. Enabling this, cluster controller will collect katalyst CNR resource into FederatedCluster if katalyst plugin was enabled in its annotations, and the scheduler will enable KatalystResourcesExist filter (it will filter out clusters without katalyst resources) and support split replicas by katalyst resources (Split priority: GPU > katalyst cpu > cpu).
* `--cluster-health-check-period=30s` The period of health check for member clusters. The minimum value is 15s.

After enabling katalyst support, if any member cluster supports katalyst, you can add the following annotation to it's fcluster to collect CNR information: `kubectl annotate fcluster kubeadmiral-member-1 kubeadmiral.io/cluster-resource-plugins="[\"node.katalyst.kubewharf.io\"]" --overwrite`


pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -91,7 +256,11 @@ func (c *FederatedClusterController) collectIndividualClusterStatus(
 	cluster = cluster.DeepCopy()
 	conditionTime := metav1.Now()
 
-	offlineStatus, readyStatus := checkReadyByHealthz(ctx, discoveryClient)
+	timeout := c.clusterHealthCheckConfig.Period / 2
```
Poor12, line 259: Origin timeout is 30s. Now it seems the default value is 15s. Is this a `Best practices` that we use half of clusterHealthCheckConfig.Period?

JackZxj: 30 seconds is actually the experience value I set before. I checked some recent statistics: normally (more than 95%) it takes no more than 10 seconds to pass the health check, and more than 10 seconds usually means that the cluster is disconnected. So I shortened it. 

On the other hand, `clusterHealthCheckConfig` is now configurable, and the minimum is currently 15 seconds (according to feedback, some people want to shorten this interval when katalyst is enabled). It feels inappropriate if the timeout is still fixed at 30s.

pkg/controllers/federatedcluster/plugins/katalyst.go
```diff
@@ -0,0 +1,110 @@
...
+
+	allocatable = make(corev1.ResourceList)
+	nodeSet := sets.New[string]()
+	for _, node := range nodes {
+		nodeSet.Insert(node.Name)
+	}
+
+	for _, uns := range unsList {
+		unsCNR, ok := uns.(*unstructured.Unstructured)
+		if !ok || !nodeSet.Has(unsCNR.GetName()) {
```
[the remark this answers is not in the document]
JackZxj: 1. Yes, CNR needs to match the nodes of the cluster, and there will be a agent runs in every katalyst enabled node which maintains it's CNR resource
2. Currently, CNR do not support taints, but the katalyst scheduler will identify the taints on native nodes, and we already have filters for these. I think the virtual node does not support the installation of katalyst's agent because its resources are too large and there is no need to distinguish between online and offline resources.

pkg/controllers/scheduler/framework/plugins/clusterresources/least_allocated.go
```diff
@@ -66,9 +66,13 @@ func (pl *ClusterResourcesLeastAllocated) Score(
 	// gpu((capacity-sum(requested))*100/capacity) * gpu_weight) / (cpu_weight + memory_weight + gpu_weight)
 	for _, resource := range resources {
 		resourceScore := leastRequestedScore(requested[resource], allocatable[resource])
-		weight := framework.DefaultRequestedRatioResources[resource]
-		score += resourceScore * weight
-		weightSum += weight
+		if weight, ok := framework.DefaultRequestedRatioResources[resource]; ok {
```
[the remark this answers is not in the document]
JackZxj: done

pkg/controllers/scheduler/framework/plugins/katalyst/exist.go
```diff
@@ -0,0 +1,63 @@
+/*
```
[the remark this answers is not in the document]
JackZxj: I felt that there might be other katalyst filters or scores, so I used katalyst as the directory. but I have renamed the files.

Poor12 (approved):

pkg/controllers/federatedcluster/clusterstatus.go
```diff
@@ -60,8 +67,166 @@ const (
...
+			hasSynceds := make([]cache.InformerSynced, 0, len(collector.hasSynceds))
+			names := make([]string, 0, len(collector.hasSynceds))
+			for gvr, synced := range collector.hasSynceds {
+				hasSynceds = append(hasSynceds, synced)
+				names = append(names, gvr.String())
+			}
+			if !cache.WaitForCacheSync(ctxWithTimeout.Done(), hasSynceds...) {
+				logger.V(4).Info("Timeout waiting for informers sync",
+					"informers", strings.Join(names, ";"))
+				return
```
limhawjia, line 210: Should we continue to collect the resource information for the gvrs that have managed to sync?

limhawjia: Oh my bad, the collection is performed by the plugin and might depend on all the  gvrs being synced

limhawjia (approved):

PR #316, opened 2024-03-19
fix: repeatedly creating and deleting mcs, the creation of derived svc occasionally fails

Poor12:
When repeatedly creating and deleting mcs, and when processing the creation event of mcs, the federatedObject of the originally derived service may still be in the deletion state. Even if the update is successful, the derived service will still be deleted and there will be no chance to retry.

### DIFF
--- a/pkg/controllers/mcs/service_import_controller.go
+++ b/pkg/controllers/mcs/service_import_controller.go
@@ -219,6 +219,33 @@ func (c *ServiceImportController) reconcile(ctx context.Context, fedObjKey fedOb
 				keyedLogger.Error(err, "Failed to sync placements to endpointSlice")
 				return worker.StatusError
 			}
+
+			// Usually derivedSvcFedObj relies on the cascade deletion mechanism of K8s for cleaning,
+			// but the GC time of K8s is not certain, so when it has not been recycled,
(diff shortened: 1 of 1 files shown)


pkg/controllers/mcs/service_import_controller.go
```diff
@@ -219,6 +219,36 @@ func (c *ServiceImportController) reconcile(ctx context.Context, fedObjKey fedOb
...
+
+			if derivedSvcFedObj != nil && derivedSvcFedObj.GetDeletionTimestamp() == nil {
+				if err = fedobjectadapters.Delete(
+					ctx,
+					c.fedClient.CoreV1alpha1(),
+					derivedSvcFedObj.GetNamespace(),
+					derivedSvcFedObj.GetName(),
+					metav1.DeleteOptions{},
+				); err != nil {
+					if apierrors.IsConflict(err) {
```
JackZxj, line 245: will delete returns a confict err? and should we ignore the NotFoundErr here?

Poor12: NotFoundErr is a normal phenomenon. It may have been deleted by gc.

JackZxj (approved):

PR #317, opened 2024-03-22
fix: retain pod spec volume when its name has default token prefix

wy-lucky:
In clusters prior to version 1.24, the default volume name mounted for service accounts is random. 
And pod spec volumes can not been updated, so we retain it.

### DIFF
--- a/pkg/controllers/sync/dispatch/retain.go
+++ b/pkg/controllers/sync/dispatch/retain.go
@@ -35,8 +35,6 @@ import (
 )
 
 const (
-	// see serviceaccount admission plugin in kubernetes
-	ServiceAccountVolumeNamePrefix = "kube-api-access-"
 	//nolint:gosec
 	DefaultAPITokenMountPath = "/var/run/secrets/kubernetes.io/serviceaccount"
 )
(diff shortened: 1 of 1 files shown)


Poor12 (approved):

mrlihanbo:
Can you found out the reference commit?

wy-lucky:
> Can you found out the reference commit?

It appears that there is no fixed name for it, and modifying the pod.spec.volumes is not allowed. Therefore, we will retain the pod.spec.volumes.

pkg/controllers/sync/dispatch/retain.go
```diff
@@ -341,23 +339,10 @@ func retainPodFields(desiredObj, clusterObj *unstructured.Unstructured) error {
...
-			}
-
-			desiredVolumes = append(desiredVolumes, nil)
-			copy(desiredVolumes[idx+1:], desiredVolumes[idx:])
-			desiredVolumes[idx] = volume
-
-			if err := unstructured.SetNestedSlice(desiredObj.Object, desiredVolumes, "spec", "volumes"); err != nil {
-				return err
-			}
+	if volumes, exists, err := unstructured.NestedFieldNoCopy(desiredObj.Object, "spec", "volumes"); err == nil &&
```
mrlihanbo, line 342: Copy the ```volumes``` filed when ```volumes``` not exist in desired object?

wy-lucky: It should always be copied.

mrlihanbo (approved):

PR #329, opened 2024-06-12
fix: delete cluster failed when failed to remove finalizers

wy-lucky:
When connecting to a member cluster using a ServiceAccountToken and removing the member cluster, the steps are as follows:

1. Obtain the connection credentials from the ServiceAccountToken secret.
2. Remove the connection credentials from the secret.
3. Remove the finalizers from the fcluster.

In the event of a failure during the third step, when re-entering the queue, the first step retries to obtain the credentials from the secret. However, since the credentials were already deleted in the second step, the attempt to retrieve the credentials fails. Therefore, in this scenario, a retry is added in the third step to address this issue.

### DIFF
--- a/pkg/controllers/federatedcluster/controller.go
+++ b/pkg/controllers/federatedcluster/controller.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	fedcorev1a1 "github.com/kubewharf/kubeadmiral/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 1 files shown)


JackZxj (approved):

pkg/controllers/federatedcluster/controller.go
```diff
@@ -457,12 +458,16 @@ func (c *FederatedClusterController) handleTerminatingCluster(
...
 	// We have already checked that we are the last finalizer so we can simply set finalizers to be empty.
-	cluster.SetFinalizers(nil)
-	if _, err := c.fedClient.CoreV1alpha1().FederatedClusters().Update(
-		ctx,
-		cluster,
-		metav1.UpdateOptions{},
-	); err != nil {
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		var err error
+		cluster, err = c.fedClient.CoreV1alpha1().FederatedClusters().Get(ctx, cluster.Name, metav1.GetOptions{ResourceVersion: "0"})
```
SOF3, line 463: Should we only try to get cluster again when we actually got a 409?

wy-lucky: PTAL.

SOF3 (approved):

PR #331, opened 2024-06-27
bugfix(RCController): opInformer add wrong eventHandler

zk-123:
i think opInformer add EventHandler  use persistOpWorker  instead of  persistPpWorker


SOF3 (approved):
Thanks! In the future, please use `fix` for the commit prefix as with [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #336, opened 2024-09-03
fix: the previous cop is completely overwritten when binding op

qclc:


pkg/controllers/override/util.go
```diff
@@ -112,12 +113,37 @@ func lookForMatchedPolicies(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get sourceObj from fedObj: %w", err)
+	}
+	return &helpData{
+		gvk:       gvk,
+		sourceObj: sourceObj,
+	}, nil
+}
+
+// 集群名称，和对应的所有patches
```
JackZxj, line 136: please translate it

pkg/util/pod/pod.go
```diff
@@ -76,3 +76,24 @@ func GetResourcePodSpec(fedObject fedcorev1a1.GenericFederatedObject, gvk schema
 	}
 	return GetPodSpec(fedObject, path)
 }
+
+func GetResourcePodSpecFromUnstructuredObj(
```
JackZxj, line 80: nit: The above func GetPodSpec can reuse this to simplify

pkg/controllers/override/util.go
qclc: sorry, this content is a comment for my own viewing only and has been deleted.

pkg/util/pod/pod.go
qclc: The common logic of this function has been extracted.

JackZxj (approved):

mrlihanbo (approved):

PR #346, opened 2024-12-16
optimize: add probe for cluster status

wy-lucky:
The cluster status may become abnormal due to a single network jitter.  If this occurs while the application is in the scheduling state, it may lead to the application not being dispatched to that cluster.  Therefore, we have optimized this by adding a detection mechanism for the cluster status to prevent unexpected scheduling results caused by a single network jitter.

### DIFF
--- a/pkg/controllers/federatedcluster/cluster_status_cache.go
+++ b/pkg/controllers/federatedcluster/cluster_status_cache.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


pkg/controllers/federatedcluster/cluster_status_cache.go
```diff
@@ -0,0 +1,103 @@
...
+	now := time.Now()
+	if saved.offlineCondition.Status != observedOfflineCondition.Status || saved.readyCondition.Status != observedReadyCondition.Status {
+		// condition status changed, record the probe timestamp
+		saved = &clusterStatusConditionData{
+			offlineCondition: observedOfflineCondition,
+			readyCondition:   observedReadyCondition,
+			probeTimestamp:   now,
+		}
+	}
+	c.update(cluster.Name, saved)
```
Poor12, line 76: Do we need to update when saved.status == observed.Status?

wy-lucky: It's only needed when saved.status == observed.Status.The code has been updated.

Poor12 (approved):

PR #353, opened 2025-01-15
feat: add resourceVersion for pod, svc and eps

Poor12:


pkg/registry/hpaaggregator/metrics/resource/pod.go
```diff
@@ -114,14 +116,24 @@ func (m *PodMetrics) pods(ctx context.Context, options *metainternalversion.List
 	}
 
 	namespace := genericapirequest.NamespaceValue(ctx)
-	pods, err := m.podLister.ByNamespace(namespace).List(labelSelector)
+	podsObj, err := m.podLister.ByNamespace(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labelSelector.String(),
```
wy-lucky, line 120: It looks like we can add resourceVersion and filedSelector here.

Poor12: PodMetrics seems not need to implement  `watcher` interface.  Keep it the same as before for now.

wy-lucky (approved):

